PR #1, opened 2018-01-19
Initial creation of the IMRT Item Ingest Service SpringBoot project. …

swilson11:
…Basic gradle setup and implementation of index endpoint.


LICENSE
```diff
@@ -0,0 +1,12 @@
+  Copyright 2017 Smarter Balanced Licensed under the
```
swilson11: I guess I should have changed this to 2018 - I just copied the file from location you specified, but didn't update it.

.gitignore
```diff
@@ -0,0 +1,5 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
+.gradle
+/build/
+!gradle/wrapper/gradle-wrapper.jar
```
millerts: Can you add the intellij related files to this git ignore?

build.gradle
```diff
@@ -0,0 +1,53 @@
+/*
```
millerts: You can remove this copyright.  The license file covers it.

build.gradle
```diff
@@ -0,0 +1,53 @@
...
+jar {
+    baseName = 'gs-spring-boot'
+    version =  '0.1.0'
+}
+
+repositories {
+    mavenCentral()
+}
+
+sourceCompatibility = 1.8
```
millerts: What do you think about using Java 9?  I have not checked if it is compatible.  But if it is I think we should use the latest.  Across the board really.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Fri Jan 19 08:30:42 PST 2018
```
millerts: If this date isn't important I'd remove this.

src/main/java/org.opentestsystem.ap.imrt.iis/IssApplication.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2018 Regents of the University of California.
```
millerts: I’m concerned you have a directory named `org.opentestsystem.ap.imrt.iis` instead of `org/opentestsystem/ap/imrt/iis`.

.gitignore
swilson11: I looked locally and I didn't find any intellij related files - I was assuming if you imported as a gradle project, that there wouldn't be any. But I can definitely add it just to be safe.

build.gradle
swilson11: I haven't spent any time investigating java 9, so I don't really have a strong opinion. A quick google search did reveal this:
https://dzone.com/articles/migrating-a-spring-boot-application-to-java-9-comp

Which sounds a little bit worrying. I'm thinking that since it was just released Sept/Oct 2017 maybe we should wait a bit longer?

millerts: Sounds good.

gradle/wrapper/gradle-wrapper.properties
swilson11: Since this is an autogenerated file I'd prefer to not edit it, though if you feel strongly I can...

PR #2, opened 2018-01-22
Add JaCoCo code coverage and findbugs support to build.gradle. Bump g…

swilson11:
…radle wrapper version to 4.4.1

Build will fail on findbugs failures
https://jira.fairwaytech.com/browse/IMRT-241


gradlew
```diff
@@ -155,7 +155,7 @@ if $cygwin ; then
 fi
 
 # Escape application args
-save () {
+save ( ) {
```
millerts: Did you change a code style here?  We should probably make sure that we use the same code style between us so that we don't end up with these type of formatting changes.  Unless this is due to the update in Gradle.

build.gradle
```diff
@@ -30,9 +30,41 @@ dependencies {
...
+ ***************************/
+tasks.withType(FindBugs) {
+    reports {
+        xml.enabled = false
+        html.enabled = true
+    }
 }
+
+// Make sure to always run the jacoco coverage report
+test.finalizedBy jacocoTestReport
```
millerts: Do we always need to run this report?  It feels like this should be an optional report to run since it will increase the build time.  I kinda feel the same way about 

* Find Bugs
* Coverage
* Checkstyle (if we use it)
* IT (Integration) tests

gradlew
swilson11: All these files are auto-generated when you run the wrapper task (./gradlew wrapper). So the changes came because I bumped the gradle version in build.gradle and re-ran the wrapper task. I was just trying to figure out how to leave comments on my pull request before starting the review to tell you that, but haven't figured it out yet...

build.gradle
swilson11: Let's chat about this in Slack.

millerts:
Still LGTM

PR #3, opened 2018-01-22
Feature/readme with setup

millerts:


README.md
```diff
@@ -1 +1,66 @@
-# AP_IMRT_ItemIngestService
\ No newline at end of file
+# IMRT Item Ingest Service
+
+The Item Ingest Service is reponsible for ingesting the item data created by the content team. This project is part of the Item Metadata and Reporting Tool.
+
+# Getting Started
```
millerts: I'm not sure if this needs to stay in the main page or should be moved to a secondary `Getting Started` page using [Github's ability to reference other markdown pages within the project](https://github.com/blog/1395-relative-links-in-markup-files).

The reason I like keeping this stuff here is because then anyone can just look at the project and know how to build.  Obviously we will not include any sensitive information.

swilson11: I'm fine with wherever, I thought earlier when you said you wanted me to create Confluence pages that was what you meant. Just let me know what you want in Confluence as well.

millerts: Sorry, I was specifically referring to our setup for CI and Bamboo.  Our CI setup will be specific for Fairway. Information that isn't dependent on any specific setup should be included within the project.  For example, if this project relies on MySQL we'll include that in this or a linked markdown.

swilson11: Sounds good. So what would you imagine going in that GettingStarted page in Confluence? A link to the README.md file in GitHub?

PR #4, opened 2018-01-24
Feature/ci setup

swilson11:


build.gradle
```diff
@@ -14,7 +16,11 @@ plugins {
     id 'findbugs'
     id 'org.springframework.boot' version '1.5.9.RELEASE'
     id 'jacoco'
+    id 'maven-publish'
+    id 'com.jfrog.artifactory' version '4.4.14'
+    id 'com.bmuschko.clover' version '2.2.0'
```
swilson11: I got this to run and it generates clover output. However, I can't get Bamboo to see the output. I wasted a bunch of time on it this morning, putting it on the backburner for now.

gradle.properties
```diff
@@ -0,0 +1,23 @@
...
+description="Web-based application for ingesting and storing assessment items."
+group=org.opentestsystem.ap
+
+version=0.1.0-SNAPSHOT
+
+# required when releasing - releaseFinish
+gitUsername=
+gitPassword=
+
+# required when publishing to Artifactory - artifactoryPublish
```
millerts: Can you add a note here and above on line 14 to the effect, **DO NOT COMMIT CREDENTIALS**.  I find that people mistakenly do it, and I think an extra warning may stop people.

Not saying you or I would do it.  Just in general.

PR #8, opened 2018-02-07
Add Kubernetes support

swilson11:
https://jira.fairwaytech.com/browse/IMRT-227

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -185,3 +185,17 @@ task dockerPushImage(type: DockerPushImage) {
 task dockerRemoveImage(type: DockerRemoveImage) {
     imageId = "${project.dockerTagBase}/${jar.baseName}:${project.version}"
 }
+
+/**
+ * Kubernetes
+ */
+task setDevKopsConfig(type: Exec) {
+    executable = 'kops'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -185,3 +185,17 @@ task dockerPushImage(type: DockerPushImage) {
...
+ */
+task setDevKopsConfig(type: Exec) {
+    executable = 'kops'
+    args = ['export', 'kubecfg', '--state', "${project.kubecfgState}", '--name', "${project.kubecfgDevName}"]
+}
+
+task updateDevKops(type: Exec) {
+    dependsOn 'setDevKopsConfig'
+    executable = 'kubectl'
+    args = ['set', 'image', 'deployment/ap-imrt-iis-deployment', "ap-imrt-iis=${project.dockerTagBase}/ap-imrt-iis:${project.version}"]
```
millerts: So we're updating each deployment with the version?  Just so I am following what is happening here.

swilson11: Updating just the dev deployment - when the config is set it sets the deployment name to be updated. I added a call to this task to the develop build on Bamboo.

PR #9, opened 2018-02-12
Rename stuff with ISS to IIS - I wish these were not so close, I have…

swilson11:
… messed this up a few times.

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/IisApplication.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/IisApplication.java
@@ -1,11 +1,11 @@
-package org.opentestsystem.ap.imrt.iss;
+package org.opentestsystem.ap.imrt.iis;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/IisApplication.java
```diff
@@ -1,11 +1,11 @@
-package org.opentestsystem.ap.imrt.iss;
+package org.opentestsystem.ap.imrt.iis;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
 @SpringBootApplication
-public class IssApplication {
+public class IisApplication {
```
millerts: What do you think of keeping the iis package but using ItemIngestService when naming the classes?

millerts: I prefer verbose naming for classes. :)

swilson11: Sounds OK to me. We probably won't use it in too many class names. I'll make the change to this one before I merge.

PR #11, opened 2018-02-14
Switch from using merge events to push events

swilson11:
Update gradle build to ensure pod is always redeployed in dev even if the docker tag does not change

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
@@ -100,7 +100,7 @@ public class GitlabClientImpl implements SystemHookListener, WebHookListener, It
     }
 
     @Override
-    public void monitorItem(Integer projectId) {
+    public void monitorItem(int projectId) {
         logger.debug("Adding project push hook id {} url {}", projectId, gitlabProperties.getWebhookUrl());
         try {
             gitLabApi.getProjectApi().addHook(projectId, gitlabProperties.getWebhookUrl(), true, false, false);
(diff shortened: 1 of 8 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -101,9 +101,9 @@ public void handleWebHook(HttpServletRequest request) {
 
     @Override
     public void monitorItem(Integer projectId) {
```
millerts: Can we tweak this to `monitorItemProject`? It lines up a little more with the variable.  Also, should this be object Integer?

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -196,7 +187,14 @@ public void onPipelineEvent(PipelineEvent pipelineEvent) {
 
     @Override
     public void onPushEvent(PushEvent pushEvent) {
-
+        logger.debug("onPushEvent ref {} branch {} path {} id {}",
```
millerts: Can we move this within the if conditional so it only logs on master pushes?  Since we don't care about anything else.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+        client.onPushEvent(event);
+        verify(itemEventListener, times(1)).onUpdateItem(projectId);
+    }
+
+
+    /**
+     * gitlab4j has a class with this name, but it is private and therefore inaccessible. Create
+     * this one so we can pass it into onProjectEvent and have it detect the correct class name
+     */
+    static final class CreateProjectSystemHookEvent extends ProjectSystemHookEvent {
```
millerts: Can this be private?

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -216,4 +214,8 @@ public GitLabApi getGitLabApi() {
     public void setGitLabApi(GitLabApi gitLabApi) {
         this.gitLabApi = gitLabApi;
     }
+
+    public void setGroup(Namespace group) {
```
millerts: Can we make this package private since it is only exposed for tests?  As long as the test and this class are in the same package we should be good.

PR #12, opened 2018-02-15
Simple rabbitMq integration

swilson11:
Create an item update queue, and place messages on it in response to Gitlab events.

### DIFF
--- a/src/main/docker/docker-compose.yml
+++ b/src/main/docker/docker-compose.yml
@@ -53,14 +53,10 @@ services:
             GITLAB_ACCESS_TOKEN:
             GITLAB_GROUP:
             GITLAB_WEBHOOK_URL:
-
         ports:
             - 8080:8080
         networks:
             - spring-cloud-network
-        links:
-            - configuration-service
-            - rabbitmq
         depends_on:
             - configuration-service
             - rabbitmq
--- a/src/main/docker/docker-compose.yml
+++ b/src/main/docker/docker-compose.yml
@@ -1,5 +1,20 @@
 version: '3'
 services:
+    rabbitmq:
+        container_name: rabbit-server
+        image: rabbitmq:management
+        ports:
+            - "5672:5672"
+            - "15672:15672"
+        networks:
+            - spring-cloud-network
+        healthcheck:
+          test: ["CMD", "rabbitmqctl", "node_health_check"]
+          interval: 30s
+          timeout: 5s
+          retries: 3
+        logging:
+            driver: json-file
     configuration-service:
         container_name: configuration-service
         image: @dockerTagBase@/configuration-service:@configServerVersion@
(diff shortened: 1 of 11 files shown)


src/main/docker/docker-compose.yml
```diff
@@ -45,8 +60,10 @@ services:
             - spring-cloud-network
         links:
             - configuration-service
+            - rabbitmq
         depends_on:
             - configuration-service
+            - rabbitmq
```
malaffoon: technically, `links` implies `depends_on` so this is redundant

malaffoon:
Just a thought since i happened to see this: if you code to spring messaging instead of rabbit/ampq, you'll avoid tying yourself to rabbit/ampq. Or at least turn it into a link-time resolution instead of having to change code.    Okay, i'll go away now.

src/main/java/org/opentestsystem/ap/imrt/iis/client/ItemEventListenerImpl.java
```diff
@@ -22,8 +26,10 @@ public void init() {
     }
 
     @Autowired
-    public ItemEventListenerImpl(ItemBankClient client) {
+    public ItemEventListenerImpl(ItemBankClient client, RabbitTemplate rabbitTemplate, Queue rabbitQueue) {
         itemBankClient = client;
+        this.rabbitTemplate = rabbitTemplate;
```
millerts: Instead of having this class know about how we send messages could we instead have a service that handles sending the message?  I understand this class is small but then we could roll a different message implementation without this class needing to change.

For example, something like [this](https://github.com/SmarterApp/TDS_ExamService/blob/develop/service/src/main/java/tds/exam/services/MessagingService.java) that could then be injected into this class.  This class knows what to do with the update and the MessageService knows how to send the messages.

malaffoon:
@millerts you can take a look at a couple places in RDW. For example, https://github.com/SmarterApp/RDW_Ingest/blob/develop/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportSource.java. That class is basically the integration between the message queue and the main app ... and it uses spring stream messaging instead of being ampq-specific. Then, by changing the compile-time import it can switch from rabbitmq to kafka (for example).

Slightly more up-to-date example: the workers in this package, https://github.com/SmarterApp/RDW_Reporting/tree/develop/report-processor/src/main/java/org/opentestsystem/rdw/reporting/processor/stream, all use simple spring stream annotations to talk to rabbitmq, without mentioning ampq by name.

swilson11:
I've generalized the implementation to try and minimize the dependence on Rabbit, but using the generalized spring AMQP stuff. I took a very quick look at the messaging stuff, my feeling at this point is that the time it would require to generalize this would be high compared to the time to write it in the first place - so I think we should defer this to such time as we do want to go to some other messaging system, and then see what makes sense at that point, especially since we don't plan to add a bunch of other queue/messaging related stuff to IIS or ISS.

PR #14, opened 2018-02-27
Add simple EventService

swilson11:
Abstract base class for code to call to generate events. Simple implementatation class that logs the events.


swilson11:
I made this a service, rather than going with static methods, in case we wanted to replace or augment simple logging with something else in the future, like kafka or some other event service.

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
```diff
@@ -0,0 +1,89 @@
+package org.opentestsystem.ap.imrt.iis.service;
+
+import org.slf4j.Logger;
+
+/**
+ * Abstract class to provide a mechanism for generating events to facilitate
+ * monitoring of the ItemIngestService.
+ */
+public abstract class EventService {
```
millerts: Why do we have an abstract class?  Do you foresee us having many different service classes that are also EventServices?

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param key  Database key for the updated item
+     * @param id  Itembank Id for the updated item
+     * @param projectId  Git projectId of the updated item
+     */
+    protected abstract void itemEvent(Logger logger, ItemEventType type, Integer key, Integer id, Integer projectId);
+
+    /**
+     * Service related events, requiring action by Operations
+     */
+    protected enum ServiceEventType {
```
millerts: Can we move the enums to be declared at the top of this class so they're easier to visually find?

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
```diff
@@ -0,0 +1,89 @@
...
+        SERVICE_WARNING, SERVICE_ERROR
+    }
+
+    /**
+     * Indicates a fatal error has occurred during processing
+     * @param logger  Logger belonging to the calling class
+     * @param format  The format string for the message
+     * @param args  A list of arguments, corresponding to the format string,
```
millerts: Can we make the Exception part of the signature?  If I'm reading the documentation correctly the Exception has to be the last Object in the args for the Exception logic to work.  I'd prefer a nullable Exception object instead.

And if we don't like forcing people to pass in null we can have two methods:
```
serviceError(Logger logger, String format, Object ... args){
  serviceError(logger, format, null, args);
}

serviceError(Logger logger, String format, , Exception e, Object ... args)
```

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventServiceLoggerImpl.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Service
+public class EventServiceLoggerImpl extends EventService {
+    @Override
+    protected void itemEvent(Logger logger, ItemEventType type, Integer key, Integer id, Integer projectId) {
+        StringBuilder builder = new StringBuilder();
+        List<Object> args = new ArrayList<>(3);
+        builder.append("ITEM EVENT type: {}");
+        args.add(type);
+        if (key != null) {
```
millerts: Rather than have these if checks can we add a helper method?  Unless you envision this to be the only three we log in this context.

```
private void append(List<Object> args, StringBuilder builder, String message, Object value){
  if(value != null) {
   builder.append(message);
   args.add(value);
  }
}
```

Then this would collapse to `append(args, builder, " key: {}", key);`

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
```diff
@@ -0,0 +1,89 @@
...
+     * Service related events, requiring action by Operations
+     */
+    protected enum ServiceEventType {
+        SERVICE_WARNING, SERVICE_ERROR
+    }
+
+    /**
+     * Indicates a fatal error has occurred during processing
+     * @param logger  Logger belonging to the calling class
+     * @param format  The format string for the message
```
millerts: I think it needs to be clear that the format is tied to slf4j logging.  For example, I don't think if I passed in `this %s is great` and then my args have `project` in it the string replace would work?  I think the format has to be `this {} is great`.

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Indicates that an item has been updated in the imrt database
+     * @param logger  Logger belonging to the calling class
+     * @param key  Database key for the updated item
+     * @param id  Itembank Id for the updated item
+     * @param projectId  Git projectId of the updated item
+     */
+    public void itemUpdatedEvent(Logger logger, int key, int id, int projectId) {
```
millerts: Can you give me an example of where this is used or will be used?

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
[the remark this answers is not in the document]
swilson11: As we discussed, changing to OperationsEventService

swilson11: Yes. I made it an interface with default methods. Not crazy about it, as now you can't make the enums and some of the methods protected, everything in the interface is public, but probably not a big deal. Apparently protected and private are coming to interfaces in java 9...

src/main/java/org/opentestsystem/ap/imrt/iis/service/EventService.java
swilson11: When a change is made to item.json, and we then update the item in the database.

PR #16, opened 2018-03-01
Add Graylog support

swilson11:


swilson11:
Merging with no review since this is identical to the ISS change.

PR #17, opened 2018-03-02
Add project lock service

swilson11:


src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+@Service
+public class ProjectLockServiceJdbcImpl implements ProjectLockService {
+    private static final Logger logger = LoggerFactory.getLogger(ProjectLockServiceJdbcImpl.class);
+
+    private final JdbcTemplate jdbcTemplate;
+    private final OperationalEventService operationalEventService;
+
+    // Maximum length of time a lock can be held.
+    @Value("${iis.maxProjectLockSeconds:60}")
+    private Integer maxProjectLockSeconds;
```
millerts: Should we have a property class that can be passed in rather than relying on field setting?  Or even just have the property passed in as a constructor?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            // The lock exists, see if it has expired
+            if (now.getTime() - existingLock.getTime() > maxProjectLockSeconds * 1000) {
+                logger.debug("Trying to lock project {}. There is an existing expired lock {}", projectId, existingLock);
+                // Lock expired, go ahead and take it. Generally this will be happen because the project
+                // has been unlocked, and the time has been set to zero
+                if (existingLock.getTime() != 0) {
+                    operationalEventService.serviceWarning(logger, null, "Existing project lock has expired for projectId {} lock time {}", projectId, existingLock);
+                }
+                int rows = jdbcTemplate.update(UPDATE_SQL, new Object[]{now, now, projectId, existingLock});
+                if (1 != rows) {
```
millerts: How can this ever be true?  The only way I can see it be true is if someone manually goes in and modifies the table directly which is something that should never be done.  Do you think we need to have logic for people directly updating and munging data?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockService.java
```diff
@@ -0,0 +1,22 @@
...
+ * Provides locking for projects.
+ */
+public interface ProjectLockService {
+
+    /**
+     * Attempts to lock the given project
+     * @param projectId  Unique identifier for the project
+     * @return  A unique identifier for the lock, which must be used in the call to unlock, or null if the lock cannot be acquired.
+     */
+    Long lockProject(int projectId);
```
millerts: I prefer us to never return null.  Can we use Java's [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) instead?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            // an exception because of the duplicate key
+            try {
+                int rows = jdbcTemplate.update(INSERT_SQL, new Object[]{projectId, now, now, now});
+                if (1 != rows) {
+                    operationalEventService.serviceError(logger, null, "Unable to create project_lock entry for projectId {}", projectId);
+                    return null;
+                }
+                return now.getTime();
+            } catch (Exception ex) {
+                return null;
```
millerts: How do you tell between an issue where someone has already locked it since we checked vs an issue with actually inserting into the DB?  If my understanding is correct this assumes that any issue inserting records means that there is already a lock.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+ * lock_time in that row is not far enough in the past to be considered expired. The epoch representation of
+ * the lock_time is used as a unique identifier for the lock. A project is considered unlocked
+ * if no row exists in the project_lock database table for the projectId, or if the row exists
+ * but the lock_time has an epoch time value of zero.
+ */
+@Service
+public class ProjectLockServiceJdbcImpl implements ProjectLockService {
+    private static final Logger logger = LoggerFactory.getLogger(ProjectLockServiceJdbcImpl.class);
+
+    private final JdbcTemplate jdbcTemplate;
```
millerts: I prefer services not to write directly to the database.  I prefer us to have repositories (or dao) be the only layer that persists data.

In general I prefer:
```
presentation layer (gitlabclient/controller) -> business layer (service) -> persistence layer (repository/dao) -> datastore
```

In my view services should have no direct knowledge of how the data is persisted.  For example, we're using the database here, but what if I want to use redis in the future?  I'd have to modify this class directly.  However, if I had a repository interface I could roll a new implementation that has the same behavior yet the business logic in here (handling locking, what happens if it fails) would not change.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void unlockProject(int projectId, long lockId) {
+        Date now = new Date();
+
+        // Update the row in the database to set the lock_time back to zero, but only if the existing lock_time
+        // matches the supplied lock value
+        int rows = jdbcTemplate.update(UPDATE_SQL, new Object[]{new Date(0), now, projectId, new Date(lockId)});
+        if (1 != rows) {
```
millerts: Shouldn't this conditional be `1 >= rows`?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    @Override
+    public void unlockProject(int projectId, long lockId) {
+        Date now = new Date();
+
+        // Update the row in the database to set the lock_time back to zero, but only if the existing lock_time
+        // matches the supplied lock value
+        int rows = jdbcTemplate.update(UPDATE_SQL, new Object[]{new Date(0), now, projectId, new Date(lockId)});
```
millerts: Why update rather than delete?  So if I have a million items there will be a million lock records in this table.

src/test/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+                "  lock_time  TIMESTAMP NOT NULL,\n" +
+                "  created_at TIMESTAMP NOT NULL DEFAULT current_timestamp,\n" +
+                "  updated_at TIMESTAMP NOT NULL DEFAULT current_timestamp,\n" +
+                "  updated_by VARCHAR     NOT NULL,\n" +
+                "  PRIMARY KEY (project_id)\n" +
+                ")");
+    }
+
+    @Test
+    public void testProjectLock() {
```
millerts: In general I prefer separate unit tests for each logic branch being tested. As you have it here each time I change any logic branch I will have to update this single method.  Basically any comment you have within this method should be a separate test.

I also would like us to discuss test names.  For example, reading this method visually does not indicate to me what is being tested.  I prefer something like `shouldLockProject` for the happy path. Taking my comment above I would like to see a `shouldHandleMultipleLockRecordsForProject` for the logic branch that handles when there is more than 1 record for a specific project.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockService.java
millerts: Or should this simply return boolean?  As far as I can tell this is used to see if it worked or not?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ProjectLockServiceJdbcImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        if (1 != rows) {
+            operationalEventService.serviceError(logger, null,"Unable to unlock project for projectId {} lock {}", projectId, lockId);
```
millerts: Should this service log or should something at a higher level log?  What would the calling service do if it was unable to secure the lock?  I feel like we should throw here (runtime) rather than log allowing the client to handle the Exception or let it bubble and be caught by the general Exception handler.

PR #18, opened 2018-03-07
On an item update, retrieve item.json from GitLab and parse it into a…

swilson11:
… Java Object

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
@@ -3,7 +3,7 @@ package org.opentestsystem.ap.imrt.iis.service;
 import org.gitlab4j.api.models.RepositoryFile;
 import org.opentestsystem.ap.common.model.Item;
 import org.opentestsystem.ap.imrt.iis.client.ItemBankClient;
-import org.opentestsystem.ap.imrt.iis.util.ItemUtil;
+import org.opentestsystem.ap.imrt.iis.util.ItemParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.amqp.rabbit.annotation.RabbitListener;
@@ -31,10 +31,13 @@ public class ItemUpdateService {
 
     private final ProjectLockService projectLockService;
     private final ItemBankClient itemBankClient;
+    private final ItemParser itemUtil;
 
-    public ItemUpdateService(ProjectLockService projectLockService, ItemBankClient itemBankClient) {
+    public ItemUpdateService(ProjectLockService projectLockService, ItemBankClient itemBankClient,
+                             ItemParser itemUtil) {
         this.projectLockService = projectLockService;
         this.itemBankClient = itemBankClient;
+        this.itemUtil = itemUtil;
     }
 
     /**
@@ -77,7 +80,7 @@ public class ItemUpdateService {
             logger.debug("Retrieved item.json from gitlab commit id {}", itemFile.get().getLastCommitId());
             byte json[] = Base64.getDecoder().decode(itemFile.get().getContent());
             // This will throw a runtime exception if there is a problem converting the item
-            Item item = ItemUtil.readItem(json);
+            Item item = itemUtil.readItem(json);
             logger.debug("converted item id {} type {}", item.getId(), item.getType());
             // TODO store to DB
         } else {
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
@@ -47,7 +47,7 @@ public class ItemUpdateService {
      * @param itemId  Unique item bank id for the item that has been updated.
      */
     @RabbitListener(queues = "item-update")
-    public void processMessage(Integer itemId) {
+    public void updateItem(Integer itemId) {
         logger.debug("Received item update message for id {}", itemId);
 
         Optional<Long> lock = Optional.empty();
@@ -60,7 +60,7 @@ public class ItemUpdateService {
                 throw new RuntimeException("Project is locked, requeue message");
             }
             // We got the lock, so process the update
-            processItemUpdate(itemId);
+            performItemUpdate(itemId);
         } finally {
             if (lock.isPresent()) {
                 projectLockService.unlockProject(itemId, lock.get());
@@ -73,7 +73,7 @@ public class ItemUpdateService {
      * parse it, and write it to our repository.
      * @param itemId  Unique git id for the item to be processed.
      */
-    private void processItemUpdate(int itemId) {
+    private void performItemUpdate(int itemId) {
         // See if we can get a file. For now we will just grab latest on master
         Optional<RepositoryFile> itemFile = itemBankClient.getFile("item.json", itemId, "master");
         if (itemFile.isPresent()) {
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
@@ -1,14 +1,21 @@
 package org.opentestsystem.ap.imrt.iis.service;
 
+import org.gitlab4j.api.models.RepositoryFile;
+import org.opentestsystem.ap.common.model.Item;
+import org.opentestsystem.ap.imrt.iis.client.ItemBankClient;
+import org.opentestsystem.ap.imrt.iis.util.ItemUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.amqp.rabbit.annotation.RabbitListener;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Service;
 
+import java.util.Base64;
+import java.util.Optional;
+
 /**
- * Sample class for listening on the ite update queue - just meant as a proof
- * of concept. I'm not crazy about having to hardcode the queue name in the
+ * Service responsible for listening on the ite update queue and processing item update
+ * messages. I'm not crazy about having to hardcode the queue name in the
  * annotation...
  *
  * Checking on the spring default configuration, it appears to be OK for the purpose of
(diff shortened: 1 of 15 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
```diff
@@ -22,8 +29,60 @@
...
+        Optional<RepositoryFile> itemFile = itemBankClient.getFile("item.json", itemId, "master");
+        if (itemFile.isPresent()) {
+            logger.debug("Retrieved item.json from gitlab commit id {}", itemFile.get().getLastCommitId());
+            byte json[] = Base64.getDecoder().decode(itemFile.get().getContent());
+            // This will throw a runtime exception if there is a problem converting the item
+            Item item = ItemUtil.readItem(json);
+            logger.debug("converted item id {} type {}", item.getId(), item.getType());
+            // TODO store to DB
+        } else {
+            logger.debug("No item.json is currently available for project id {}", itemId);
```
millerts: I don't think this should ever be possible.  If we are seeing things to merge to master and we aren't seeing an item.json I think we should probably either throw here and let the ExceptionHandler log it or log it here as a service warn/error.

I realize there are items in master today that do not have item.json but I think this needs more visibility as I think the new way forward is to always use IAT

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
```diff
@@ -0,0 +1,33 @@
...
+import org.opentestsystem.ap.common.model.Item;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * Utility class for item related methods
+ */
+public class ItemUtil {
```
millerts: Can we rename this to ItemReader or ItemMapper?

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    /**
+     * Parse an item from a byte array
+     * @param json  Byte array containing the json represenation of the item.
+     * @return  Item object.
+     */
+    public static Item readItem(byte[] json) {
```
millerts: I don't think anything in this class should be static.  If we want hide the way we map the item data from classes that need to leverage the item data file directly I think they should just have this class be injected.   

I also feel the object mapper should also be injected via Spring.  That way everything in the entire IIS system will use the same object mapper with all the provided modules.  SpringBoot provides one for free with Spring web boot or spring web mvc (I can't remember).

src/main/resources/application.yml
```diff
@@ -1,8 +1,14 @@
...
+# when using the remote configuration service.
 spring:
   jpa:
     hibernate:
       ddl-auto: validate
     database-platform: org.hibernate.dialect.PostgreSQLDialect
+  datasource:
+    url: "jdbc:postgresql://localhost:5432/imrt"
+    username: imrt_ingest
+    password: password
```
millerts: Please remove username/password information.  I realize this is for local but I'd like to get in the habit of never having it in files in github.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -15,7 +17,11 @@
 import org.opentestsystem.ap.imrt.iis.config.ItemBankProperties;
 import org.opentestsystem.ap.imrt.iis.service.OperationalEventService;
 
+import java.util.Optional;
+
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
millerts: Sorry, I didn't see this earlier but we may want to think about leveraging [assertj](http://joel-costigliola.github.io/assertj/) or [fest assertions](https://github.com/alexruiz/fest-assert-2.x).  The assertions are pretty powerful and read better imho.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
```diff
@@ -22,8 +29,60 @@
...
+        try {
+            lock = projectLockService.lockProject(itemId);
+            if (!lock.isPresent()) {
+                logger.debug("Project {} is already locked. Requeue message for later", itemId);
+                throw new RuntimeException("Project is locked, requeue message");
+            }
+            // We got the lock, so process the update
+            processItemUpdate(itemId);
+        } finally {
+            if (lock.isPresent()) {
```
millerts: If the lock isn't present on line 55 it'll throw making this conditional impossible to hit, correct?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
```diff
@@ -22,8 +29,60 @@
...
+        this.itemBankClient = itemBankClient;
+    }
+
+    /**
+     * This method is invoked to process messages on the item-update queue.
+     * If the method returns normally, the message will be acked. If the method
+     * throws an exception, the message will be requeued for later processing.
+     * @param itemId  Unique item bank id for the item that has been updated.
+     */
     @RabbitListener(queues = "item-update")
```
millerts: Can you tell me your thinking to combine listening to the message, grabbing a lock, and processing an item all in a single service?

src/main/resources/application.yml
swilson11: I'm having trouble figuring out the best way to set this up. Locally, there are 2 modes of execution - one for the tests, where we use the "test" db and we need a user will more privileges, and one where we use the real db (both local). If I set them in environment variables, then I can run both iis and the test from the same terminal window without redefining the environment variables, which is a bit of a pain. Any suggestions?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
swilson11: This is a finally clause, it will always be executed, even if an exception if thown. I want to be sure we always unlock, even if say something in processItemUpdate throws a RuntimeException.

millerts: True dat. My bad.  ignore my stupid comment.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
swilson11: Probably easier if you just tell me how you'd like to see them divided up.

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
swilson11: OK, so just to make sure I am clear, you want this class to be a Service, called ItemXxx (maybe ItemParser, since it seems like it will just have this one method), and then Autowire in the object mapper?

millerts: I don't think it needs to be a service.  I think it can be a class and we can leverage the `@Component` annotation so Spring will add it to the container.  It can have a constructor that takes in an autowired object mapper.  Ideally, we will have one object mapper for the entire project.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
millerts: Understood.  Unfortunate.

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
swilson11: Do you envision us parsing anything else besides Items in iis?

src/main/resources/application.yml
millerts: Can we not create a separate environment variables for tests?

swilson11: You can do it inside the IDE. But not from the command line. But I can just use 2 different tabs in the terminal window.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
swilson11: I haven't used either of those, have you used them before? Which one do you think is better?

millerts: I know other projects use assertj.  I'm not asking for you to update your classes now.  I just wanted to offer it as something to leverage going forward.

swilson11: I was just reading about it, looks like it could be useful. Especially if it means you don't have to implement equals methods on objects you want to compare during tests!

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
millerts: For phase 1, no. I can't foresee what will happen in the future of IMRT past what is in phase 1-3 which right now is all about parsing an item.  I suggested the interface because any data that is ingested into IMRT from gitlab will require a parser as it is defined here.

src/main/resources/application.yml
millerts: You can leave the datasource defined as is here. I would like us not to commit the values for username/password in Github since this is a public repo.

src/main/java/org/opentestsystem/ap/imrt/iis/util/ItemUtil.java
swilson11: So what do you see as the advantage of using the interface? Not pushing back, just feel like I'm missing something.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateService.java
millerts: What do you think about renaming this method `updateItem` and change the private method to some other name? 

As for my other question please just ignore it.

PR #19, opened 2018-03-08
Hibernate

swilson11:


src/main/java/org/opentestsystem/ap/imrt/iis/service/DomainManager.java
```diff
@@ -0,0 +1,15 @@
+package org.opentestsystem.ap.imrt.iis.service;
+
```
swilson11: I didn't actually mean to include this class, sorry. Once the Entity classes are all done, the idea is that this class would provide a transactional method for taking an item from the itembank, along with it's corresponding git commit details, and create all the connected entities (ImrtItem, ItemGit, standardId, StimLink if applicable, and eventually itemLog) and save them to the DB in a single transaction.

src/main/java/org/opentestsystem/ap/imrt/iis/model/BaseEntity.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Base class for all database entities. Provides common fields and
+ * functionality
+ */
+@TypeDefs({
+        @TypeDef(name = "json", typeClass = JsonStringType.class),
+        @TypeDef(name = "jsonb", typeClass = JsonBinaryType.class)
+})
+@MappedSuperclass
+public class BaseEntity {
```
millerts: Abstract?

src/main/java/org/opentestsystem/ap/imrt/iis/model/BaseEntity.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Column(nullable = false)
+    @Temporal(TemporalType.TIMESTAMP)
+    @LastModifiedDate
+    private Date updatedAt;
+
+    @NotBlank
+    private String updatedBy;
+
+    public Date getCreatedAt() {
```
millerts: Can we use [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) rather than Date?  Or some other java.time object as they are replacements for the Java's previous date implementations.

Same comment for any field that uses Java's Date.

src/test/java/org/opentestsystem/ap/imrt/iis/BaseIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+        jdbcTemplate.execute("DELETE FROM item");
+        jdbcTemplate.execute("TRUNCATE TABLE project_lock");
+    }
+
+    /**
+     * Generate a fully populated ImrtItem to use in testing
+     * @return  A fully populated ImrtItem
+     */
+    protected ImrtItem generateItem() {
+        Item item = new StimItem("testId");
```
millerts: Rather than use this style I suggest we use the [test builder pattern](http://www.natpryce.com/articles/000714.html)

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemGit.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+import javax.persistence.UniqueConstraint;
+
+@Entity
+@Table(name = "item_git")
+@EntityListeners(AuditingEntityListener.class)
+public class ItemGit extends BaseEntity {
```
millerts: Can we tweak this a bit to something like `ItemGitInformation`?  I realize you're naming the classes like the tables but I think this class would benefit from having a more declarative name.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ImrtItem.java
```diff
@@ -0,0 +1,143 @@
...
+    private String grade;
+
+    @NotBlank
+    private String workflowStatus;
+
+    @NotBlank
+    private String itemType;
+
+    @NotBlank
+    private String dok;
```
millerts: Can we use `depthOfKnowledge`?

src/test/java/org/opentestsystem/ap/imrt/iis/BaseIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Autowired
+    protected JdbcTemplate jdbcTemplate;
+
+    @Autowired
+    protected MockMvc mvc;
+
+    @Before
+    public void setup() {
+        // Before each test, empty out the tables in the test db.
+        jdbcTemplate.execute("DELETE FROM item_git");
```
millerts: We should never need to do this.  All our DB related tests should use the `@Transactional` and so that the data is never committed during each run.

build.gradle
```diff
@@ -115,6 +116,11 @@ findbugs {
  */
 task unitTest (type: Test) {
     exclude '**/*IntegrationTest.class'
+    exclude '**/*RepositoryTest.class'
```
millerts: Can we have a single way to name our integration tests?  Other Smarter Balanced projects use IT or IntegrationTest.  Or at least TDS and RDW do that style.

src/test/java/org/opentestsystem/ap/imrt/iis/BaseIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Base class for all spring boot integration tests, including repository tests.
+ * If all tests extend this base class, they will all use the same spring context,
+ * which will make the tests run much faster. Just be sure not to dirty up the
+ * context in ways that will affect other tests
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+public class BaseIntegrationTest {
```
millerts: Abstract?

src/test/java/org/opentestsystem/ap/imrt/iis/BaseIntegrationTest.java
millerts: If the test database has data in it someone either did that manually or there is a test that isn't using the annotations properly for db tests.

src/main/java/org/opentestsystem/ap/imrt/iis/model/BaseEntity.java
swilson11: Nice. Had to pull in some updated hibernate dependencies to get support for this, apparently it's not official until hibernate 5.2, which gets pulled in once we go to spring boot 2.0. Maybe we should consider upgrading the spring boot version, but that seems like it might not be trivial.

millerts: I'm up for going to the latest stable released version.  Though we may want to wait until phase 2 for that work if there is a lot of refactor required.

build.gradle
swilson11: I'd prefer to keep them separate. If at some point, we can use an embedded DB for Postgres, then we can make these repository tests not require the full spring context, and take them out of the IntegrationTests bucket. If we keep the current naming strategy, this would be a one line change in the gradle file, instead of renaming all the test classes.

millerts: Ok. If we want both I think both should have the word Integration in them. I would be surprised if I came into a Java project and had integration tests without the word integration in the test name.

swilson11: If I put the word integration in the name then I can't easily separate them in gradle.

millerts: We can discuss this offline.

src/main/java/org/opentestsystem/ap/imrt/iis/model/BaseEntity.java
swilson11: Spring 2 went GA on March 1st. I think we should do some investigation before trying to switch - maybe you could create a Jira Item for it.

PR #26, opened 2018-03-21
Add logic to process each commit for an item

swilson11:


build.gradle
```diff
@@ -49,8 +49,8 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-common:0.3.26'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.3'
+    compile 'org.opentestsystem.ap:ap-common:0.3.28'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.4-SNAPSHOT'
```
swilson11: will update once common build is done.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -225,15 +229,37 @@ void setGroup(Namespace group) {
...
     @Override
-    public Optional<RepositoryFile> getFile(String filePath, Integer projectId, String ref) {
+    public RepositoryFile getFile(String filePath, int itemId, String ref) {
         try {
-            RepositoryFile file = gitLabApi.getRepositoryFileApi().getFile(filePath, projectId, ref);
-            return Optional.of(file);
+            return gitLabApi.getRepositoryFileApi().getFile(filePath, itemId, ref);
         } catch (GitLabApiException e) {
-            // This could happen for a number of reasons, most likely being that the requested
-            // file does not exist.
```
swilson11: Now that we are calling this with a commit ID associated with item.json, we don't get the not found errors as exceptions any more.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -0,0 +1,193 @@
+package org.opentestsystem.ap.imrt.iis.service;
+
```
swilson11: This was a rename of ItemUpdateService, not sure why it is showing up as a new file and a delete...

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Synchronize an item to the item bank, by ingesting any revisions we have not
+     * already processed.
+     *
+     * @param itemId Unique git id for the item to be processed.
+     */
+    private void syncItemToItemBank(int itemId) {
+        try {
+            // See if we have any existing itemGitInformation for this item in the database
+            ItemGitInformation existingGitInfo = itemGitInformationRepository.findOne(itemId);
```
swilson11: I started down the path of adding a method to return an Optional for this, but I found out that passing an optional around to other methods (which is mainly what I do with the existingGitInfo) causes all kinds of dire compiler warnings. Any thoughts or suggestions? Maybe we can talk about it tomorrow.

src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swilson11: When we added log4j2-spring.xml for Graylog, it killed console logging in intellij during test execution. Adding this in to bring it back.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -0,0 +1,193 @@
...
+    private ItemGitInformation syncItemRevision(int itemId, ItemGitInformation existingGitInfo, Commit commit) {
+        logger.debug("Synchronizing commit: date {} id {} name {} email {}", commit.getCommittedDate(), commit.getId(),
+                commit.getCommitterName(), commit.getCommitterEmail());
+
+        // Pull down item.json for this commit
+        RepositoryFile itemFile = itemBankClient.getFile("item.json", itemId, commit.getId());
+        logger.debug("Retrieved item.json from gitlab commit id {}", itemFile.getLastCommitId());
+
+        // Create an ItemRevision
+        ItemRevision revision = new ItemRevisionImpl(itemId, existingGitInfo, commit, itemFile, itemParser);
```
millerts: Per side discussion:

Another potential idea is to modify the `ItemRevisionWriter` to have three methods:

```
interface ItemRevisionService {
  Optional<ItemRevision> getItemRevision(itemId, ItemGitInformation);
  Collection<ItemRevision> getItemRevisions(itemId, itemGitInfomration);
  saveItemRevisions(itemRevisions)
}
```

The get method would do the following (the list would just do it for all commits):

1. Get the item from git
2. Convert it to the ItemRevision

The `ItemRevisionService` would interact with `ItemBankClient` to get both the commits and the items for each commit.

**Note** - This captures the discussion but it was not determined that this is necessary.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -225,15 +229,37 @@ void setGroup(Namespace group) {
...
+                    itemId, "master", Date.from(since), null, "item.json");
+            // Gitlab returns the list newest first, we want oldest first
+            Collections.reverse(commits);
+            return commits;
+        } catch (GitLabApiException e) {
+            // If there are no commits for item.json, then the above call returns an empty list.
+            // So any exception means something went wrong
+            operationalEventService.serviceError(logger, e, "Unable to retrieve history for itemId {} since {}",
+                    itemId, since);
+            throw new RuntimeException(e);
```
millerts: Per side discussion:

We may want to consider creating our own typed RuntimeException for certain conditions.  For example, we may want to throw an `ItemBankException` whenever we have issues connecting/retrieving data from the itembank.  And maybe a different RuntimeException for other generic item processing errors rather than just throwing Runtime.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -36,14 +41,17 @@
...
-        this.gitId = gitId;
-        this.gitPath = gitPath;
-        this.item = item;
-        this.revisionId = revisionId;
-        this.revisionDate = revisionDate;
-        this.author = author;
+    public ItemRevisionImpl(int itemId, ItemGitInformation existingItemGitInformation, Commit commit,
+                            RepositoryFile itemFile, ItemParser itemParser) {
+        // Parse out the item
+        byte json[] = Base64.getDecoder().decode(itemFile.getContent());
```
millerts: Per side discussion:

Based on the changes in ItemBankClient lines 47-49 should no longer be necessary in this POJO since ItemBankClient will return an item in its method return.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -51,7 +55,58 @@
...
+        when(api.getCommits(99, "master", Date.from(since), null, "item.json"))
+                .thenThrow(ex);
+        when(gitLabApi.getCommitsApi()).thenReturn(api);
+
+        try {
+            client.getItemHistory(99, since);
+            fail();
+        } catch (RuntimeException re) {
+            assertThat(ex).isEqualTo(re);
+            verify(api, times(1)).getCommits(
```
millerts: I do not think you need the `times(1)` here because [verify](http://static.javadoc.io/org.mockito/mockito-core/2.16.0/org/mockito/Mockito.html#verify-T-) defaults to checking for a single execution.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -73,8 +128,12 @@ public void errorOnGetFile() throws GitLabApiException {
         when(api.getFile("test", 55, "master")).thenThrow(e);
         when(gitLabApi.getRepositoryFileApi()).thenReturn(api);
 
-        Optional<RepositoryFile> result = client.getFile("test", 55, "master");
-        assertFalse(result.isPresent());
+        try {
+            client.getFile("test", 55, "master");
+            fail();
+        } catch (RuntimeException ex) {
```
millerts: If the comment about creating certain typed Exceptions I think you may be able to remove the try/catch and just have `@Test (expected = ItemBankException.class)` which will fail if the Exception isn't called.

swilson11: You can do that, however then no code after the method call that throws the exception will be executed. So it you just want to verify the contents of the exception you can use something like thrown.expected, but I don't know of any way to use verify in this case.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
swilson11: I'll do this as part the changes to add the business logic framework and/or item_log entries

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
swilson11: I'm going to think about this one as I make the changes to add the business logic processing.

swilson11:
Thanks. It shows up for me - maybe just needed to refresh your page? I have
done that to myself a few times with git. Hopefully I didn't do something
odd when I pushed the commit, I just used intellij...

PR #29, opened 2018-03-23
Feature/business logic

swilson11:


src/main/java/org/opentestsystem/ap/imrt/iis/business/ItemRevisionBusinessRule.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.ap.imrt.iis.business;
+
+import org.opentestsystem.ap.imrt.common.model.ItemGitInformation;
+import org.opentestsystem.ap.imrt.iis.model.ItemRevision;
+
+/**
+ * A business rule to be applied as an item revision is being ingested
+ */
+public interface ItemRevisionBusinessRule {
```
millerts: What do you think about the name `ItemRevisionProcessor`?  And the method can be captureItemDelta() or something like that?

src/main/java/org/opentestsystem/ap/imrt/iis/exception/GitLabApiRuntimeException.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Wrapper to create a RuntimeException from a {@link org.gitlab4j.api.GitLabApiException}
+ */
+public class GitLabApiRuntimeException extends RuntimeException {
+
+    /**
+     * @param gitLabApiException exception cause
+     */
+    public GitLabApiRuntimeException(GitLabApiException gitLabApiException) {
```
millerts: Are the GitLabApiExceptions clear in why something failed?  I'm wondering if there is any value in providing an additional `String message` argument in here.  You know more about the gitlab api than me.

src/main/java/org/opentestsystem/ap/imrt/iis/business/ItemRevisionBusinessRule.java
millerts: Or maybe just remove the word "Business" from both the class and the method?

PR #30, opened 2018-03-24
Feature/rabbit exception handling

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/config/ItemUpdateWorkQueue.java
```diff
@@ -29,4 +36,21 @@ TopicExchange exchange() {
     Binding binding(Queue queue, TopicExchange exchange) {
         return BindingBuilder.bind(queue).to(exchange).with(queueName);
     }
+
+    @Bean
+    SimpleMessageListenerContainer itemUpdateListenerContainer(final ConnectionFactory connectionFactory,
+                                                                   final OperationalEventService operationalEventService,
+                                                                   final ItemUpdateMessageListener itemUpdateMessageListener,
+                                                                   final ItemIngestServiceProperties properties) {
+        final SimpleMessageListenerContainer container = new SimpleMessageListenerContainer();
```
millerts: I chose to go with this pattern I've used before in a previous project.  This setup will retry messages as configured within the properties allowing time (configured) to pass between retrying the message.  After `n` times (configurable) it will simply send it to the recoverer and allow it to be logged.

If we decide this is too simple and should have a dead letter queue we can look at this again.

src/main/java/org/opentestsystem/ap/imrt/iis/message/ItemUpdateMessageListener.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Handles processing of item update message listener
+ */
+@Component
+public class ItemUpdateMessageListener {
```
millerts: I pulled the message handling out to allow for some additional logging and allowing the ItemUpdateNotificationHandler to focus on processing the update message and not have to deal with error handling.  The handler now will let Exceptions bubble and will be logged here.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -28,7 +26,7 @@
  * are kept on the queue until there is a free worker.
  */
 @Service
-@ConditionalOnProperty(value = "test.enabled", havingValue = "false", matchIfMissing = true)
+//@ConditionalOnProperty(value = "test.enabled", havingValue = "false", matchIfMissing = true)
```
millerts: I commented this out for now.  This was making tests fail locally, and I need to spend a little more time understanding the reason we're leveraging this property.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -58,8 +56,6 @@ public ItemUpdateNotificationHandler(ProjectLockService projectLockService, Item
      *
      * @param itemId Unique item bank id for the item that has been updated.
      */
-    // TODO update comment with link to look at rabbit config
-    @RabbitListener(queues = "item-update")
```
millerts: Moved to the listener class.  I was looking at the Spring documentation, but it looks like we would have to move to the 2.0 version of the project to leverage the annotation error handling.  If I'm incorrect please let me know.  Here is the [link](https://docs.spring.io/spring-amqp/docs/current/reference/html/_reference.html#annotation-error-handling) to the documentation.

src/main/java/org/opentestsystem/ap/imrt/iis/config/ItemUpdateWorkQueue.java
```diff
@@ -29,4 +36,21 @@ TopicExchange exchange() {
...
+        container.setConnectionFactory(connectionFactory);
+        container.setQueueNames(queueName);
+        container.setMessageListener(new MessageListenerAdapter(itemUpdateMessageListener, "handleMessage"));
+        container.setAdviceChain(RetryInterceptorBuilder.stateless()
```
millerts: This advice chain is the magic.  If we start having a bunch of queues this may need to be revisited but right now I foresee at least for phase 1 a single queue to handle updates.

src/main/java/org/opentestsystem/ap/imrt/iis/message/ItemUpdateMessageListener.java
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @param itemId the item id that needs to be updated
+     */
+    public void handleMessage(final Integer itemId) {
+        logger.debug("Received item update message for item id: {}", itemId);
+
+        try {
+            itemUpdateNotificationHandler.processItemMessage(itemId);
+        } catch (Throwable e) {
+            operationalEventService.serviceWarning(logger, e, "Unexpected error when processing item update for item %d", itemId);
```
millerts: I chose to go with warning here because the message will be retried `n` times.  The recoverer class logs at `error` because if it reaches that class the message cannot be processed.

src/main/java/org/opentestsystem/ap/imrt/iis/message/ItemUpdateMessageRecoverer.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ItemUpdateMessageRecoverer.class);
+    private final OperationalEventService operationalEventService;
+
+    public ItemUpdateMessageRecoverer(final OperationalEventService operationalEventService) {
+        this.operationalEventService = operationalEventService;
+    }
+
+    @Override
+    public void recover(final Message message, final Throwable cause) {
+        operationalEventService.serviceError(logger, cause, "Unable to process update message: %s", message);
```
millerts: This is only called if the message fails more than the configured threshold. By the time it reaches this class the message cannot be processed by our system and therefore an error log is created.

millerts:
Merging.  @jeffjohnson9046 , please feel free to add comments and I'll follow up with another PR to correct issues.

PR #34, opened 2018-04-05
Rename itemId to projectId where appropriate

millerts:
During a debugging session with @jeffjohnson9046 we both came to the realization that at certain parts of ingest the `itemId` actually referred to the item's git project id.  This was confusing to Jeff and I since we don't know the internals of the system as well as others.

This has been mentioned in other discussions.


src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -240,7 +240,7 @@ void setGroup(Namespace group) {
     }
 
     @Override
-    public List<ItemBankItemRevision> getItemHistory(int itemId, Instant since) {
+    public List<ItemBankItemRevision> getItemHistory(int projectId, Instant since) {
```
millerts: I used `projectId` everywhere but I'm not married to the name.  I think it should not be `itemId` as that means the item id as created via IAT.  This is the unique id used by gitlab to track projects.

swilson11: I agree this is confusing. I think it would be worthwhile to sit together and take a look at the naming for all the database tables, fields and identifiers. In the short term, I'd prefer something that wasn't gitlab specific, maybe like item bank id? But not a big deal if you want to leave it at project id.

millerts: Sure, we can discuss it today's standup.  Just my quick opinions:

* itemId should always refer to the item id as stored within the item json.  This is what the client refers to it as well.
* itemBankId (or bankId) should be used to refer to the internal Gitlab project id when used.  This has a particular meaning in the TDS world which is why Jeff and I didn't use it last night, but it is probably more accurate.
* Imrt's items shouldn't need to use key and item_key where appropriate.

PR #35, opened 2018-04-05
IMRT-300: Feature/item sync job

jeffjohnson9046:
[IMRT-300](https://jira.fairwaytech.com/browse/IMRT-300):  Add the ability to walk the Item Bank and add monitoring for any un-monitored item
[IMRT-298](https://jira.fairwaytech.com/browse/IMRT-298):  Update the IIS Sync Job to add updating any item that is out of sync

* Implement Spring Batch:
  * Only one instance of the job will run at a time
  * Job will run asynchronously, called via a web endpoint and returning to the caller immediately
  * Job is repeatable - the same job can be run multiple times

Unfortunately, right now the item synchronization job is one step that walks the entire item bank in GitLab.  I realize this is not an ideal configuration; the job is "opaque" (i.e. it is hard to tell how many items got processed, monitor progress, etc.)  The team has discussed being able to see each item as it gets processed.  While this PR is under review, I will create a new story and a subsequent PR to break the item synchronization job into smaller, more granular steps that provide greater transparency into what's happening.

Also, synchronization - like calculator - is surprisingly hard to spell and never looks quite right.


src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    @Autowired
+    public ItemSynchronizationServiceImpl(final ItemBankClient itemBankClient,
+                                          final ItemEventListener itemEventListener) {
+        this.itemBankClient = itemBankClient;
+        this.itemEventListener = itemEventListener;
+    }
+
+    @Override
+    public void synchronize() {
+        final List<Integer> projectIds = itemBankClient.getProjectIds(100);
```
jeffjohnson9046: This is what I was referring to in the summary of this PR; one "step" that processes all the projects in the item bank.  It'd be better if the `synchronize()` method took in a project id and worked against that single id (and maybe a "since" date?).  Some sort of reader class could read in all the project ids in GitLab, then feed those into this processor for... well... processing.

I know that's what we've talked about (several times), but I unfortunately just ran out of time.  As I said - I'll create a story and whoever works it can refactor to give us more transparency into what the job's doing.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+        for (final Integer projectId : projectIds) {
+            // If the project does not any webhooks in place or if the project does not have any webhooks that match
+            // IMRT's webhook URL configuration, create one.
+            if (!itemBankClient.isProjectMonitored(projectId)) {
+                itemEventListener.onCreateItem(projectId);
+            }
+
+            itemEventListener.onUpdateItem(projectId);
```
jeffjohnson9046: This guy right here is what handles IMRT-298

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -115,15 +121,16 @@ public void shouldGetRevisions() throws GitLabApiException, IOException {
...
     }
 
     /**
      * Validate the contents of an ItemBankItemRevision object
-     * @param revision the object to be validated
+     *
+     * @param revision           the object to be validated
      * @param expectedItemBankId the expected id
-     * @param expectedCommit commit object containing expected values
+     * @param expectedCommit     commit object containing expected values
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -79,7 +85,7 @@ public void shouldReturnEmptyList() throws GitLabApiException {
         CommitsApi api = mock(CommitsApi.class);
         Instant since = Instant.now();
         when(api.getCommits(99, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME))
-                .thenReturn(commitList);
+            .thenReturn(commitList);
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -98,7 +104,7 @@ public void shouldGetRevisions() throws GitLabApiException, IOException {
         CommitsApi api = mock(CommitsApi.class);
         Instant since = Instant.now();
         when(api.getCommits(itemBankId, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME))
-                .thenReturn(commitList);
+            .thenReturn(commitList);
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -115,15 +121,16 @@ public void shouldGetRevisions() throws GitLabApiException, IOException {
         validateItemBankItemRevision(result.get(1), itemBankId, commit2);
         validateItemBankItemRevision(result.get(2), itemBankId, commit1);
         verify(api).getCommits(
-                itemBankId, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME);
+            itemBankId, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME);
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -141,7 +148,7 @@ public void shouldThrowExceptionGetCommits() throws GitLabApiException {
         Instant since = Instant.now();
         RuntimeException ex = new RuntimeException();
         when(api.getCommits(99, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME))
-                .thenThrow(ex);
+            .thenThrow(ex);
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -160,7 +167,7 @@ public void shouldThrowExceptionGetFile() throws GitLabApiException {
         CommitsApi api = mock(CommitsApi.class);
         Instant since = Instant.now();
         when(api.getCommits(55, BRANCH_NAME, Date.from(since), null, ITEM_FILE_NAME))
-                .thenReturn(commitList);
+            .thenReturn(commitList);
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -247,7 +254,7 @@ public void shouldCreateItemOnProjectCreate() {
         gitGroup.setPath(group);
         client.setGroup(gitGroup);
         CreateProjectSystemHookEvent createEvent = new CreateProjectSystemHookEvent();
-        createEvent.setPathWithNamespace(group+"rest of path");
+        createEvent.setPathWithNamespace(group + "rest of path");
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -277,7 +284,7 @@ public void wrongGroupOnPushEvent() {
         event.setProject(project);
         event.setProjectId(projectId);
 
-         // Set the listener, but use the wrong branch. Nothing will happen
+        // Set the listener, but use the wrong branch. Nothing will happen
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

src/test/resources/application.yml
```diff
@@ -8,5 +8,4 @@ logging:
 
 spring:
   datasource:
-    url: "jdbc:postgresql://localhost:5432/test"
-
+    url: "jdbc:postgresql://localhost:5432/test"
```
jeffjohnson9046: Formatting change from IntelliJ `cmd` + `opt` + `l`, presumably

build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.28'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.7'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.9'
```
millerts: Jeff and I found an issue with an earlier refactor where git info was extending the wrong entity.  Bumped to fix.

src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationTasklet.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public ItemSynchronizationTasklet(final ItemSynchronizationService itemSynchronizationService) {
+        this.itemSynchronizationService = itemSynchronizationService;
+    }
+
+    @Override
+    public RepeatStatus execute(final StepContribution contribution, final ChunkContext chunkContext) throws Exception {
+        itemSynchronizationService.synchronize();
+
+        return null;
```
millerts: Should this return one of the RepeatStatus enums?  Finished or continued?

src/main/java/org/opentestsystem/ap/imrt/iis/config/BatchConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+ * item synchronization job should be rejected
+ * 2. The job should not run when Spring Boot starts up. Instead, the job should be started by a call to a web endpoint
+ * (e.g. a RESTful call)
+ * 3. The same job should be able to run multiple times (i.e. it should be repeatable).  That is, the Item
+ * Synchronization Process should be able to run once a day (or at some other regular interval)
+ * </p>
+ */
+@Configuration
+@EnableBatchProcessing
+public class BatchConfiguration {
```
millerts: Can we call this SpringBatchConfiguration?

src/main/java/org/opentestsystem/ap/imrt/iis/config/BatchConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+    @Autowired
+    public JobBuilderFactory jobBuilderFactory;
```
millerts: Do you need to Autowire these fields?  Since this is a `@Configuration` class Spring should just do it for you if you include it in a `@Bean` annotated method.  For example, this particular bean is used in `jobLauncher` method, and I think if you removed lines 40 & 41 everything would still work.

Same for the other autowired fields in this class.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -98,7 +99,7 @@ private void createImrtItem() {
         // Metadata may be null if this is a tutorial item
         if (metadata != null) {
             baseItem.setSubject(metadata.getSubject());
-            baseItem.setGrade(metadata.getGrade());
+            baseItem.setGrade(metadata.getGrade() == null ? StringUtils.EMPTY : metadata.getGrade());
```
millerts: Based on some items taken from the itembank we found that grade can actually be null.  When it is null it overwrites the grade's empty string default and we get constraint errors saving to the DB.

We may want to consider allowing null for more of these metadata fields.  Something that can be discussed offline at today's standup.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
millerts: Can you double check your code style settings in IntelliJ? The default is tab size 4, indent 4, and continuation indent 8.

I tried to format this file locally and don't see this change but maybe I'm not doing something right.

src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationJobExecutionListener.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+@Component
+public class ItemSynchronizationJobExecutionListener implements JobExecutionListener {
+    // The active job execution, which will act as a lock.
+    private JobExecution activeJob;
+
+    @Override
+    public void beforeJob(final JobExecution jobExecution) {
+        // Create a lock to examine the active job.  If the active job is already running, prevent the execution of
+        // the new job request.
```
swilson11: Will this work across instances - so if another instance of IIS is running the job, does this co-ordination go through the database to stop it remotely?

src/main/java/org/opentestsystem/ap/imrt/iis/client/ItemBankClient.java
```diff
@@ -33,4 +33,21 @@
      */
     List<ItemBankItemRevision> getItemHistory(int itemId, Instant since);
 
+    /**
+     * Get a collection of all {@link org.gitlab4j.api.models.Project} identifiers contained in the GitLab group.
```
swilson11: Update comment since we are no longer returning Project objects.

src/main/java/org/opentestsystem/ap/imrt/iis/client/ItemBankClient.java
```diff
@@ -33,4 +33,21 @@
...
+     *
+     * @param pageSize The number of elements the backing iterator should contain during each fetch of project ids.
+     * @return A {@link java.util.List} of project identifiers.
+     */
+    List<Integer> getProjectIds(int pageSize);
```
swilson11: I think the fact that we are using paging should be hidden inside the implementation class, rather than passing in a parameter here. I think we can always just use the max size, which is 100, and make this a constant inside GitLabClientImpl

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemSynchronizationController.java
```diff
@@ -0,0 +1,29 @@
...
+    public ItemSynchronizationController(final JobLauncher jobLauncher,
+                                         final Job job) {
+        this.jobLauncher = jobLauncher;
+        this.job = job;
+    }
+
+    @PostMapping("/sync")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void synchronize() throws Exception {
+        jobLauncher.run(job, new JobParameters());
```
swilson11: I was thinking that here we would check for an instance of a running job in the database (which might be running on another instance). If there was one, we would return an error from this API, so that we never started 2 jobs at once. I think the spring batch framework provides some APIs for checking if a job is already in progress. This is something we can add in another JIra though, just wanted to make sure we captured it somewhere.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
jeffjohnson9046: Sure enough - my continuation indent was also set to 4.  I've updated my IntelliJ settings.

src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationJobExecutionListener.java
jeffjohnson9046: That's a good question - I'll have to do some research to determine if the `JobExecution` comes from the database.  If it does, then I think we're fine.  I'd like to think the `JobExecution` passed into `beforeJob` comes from the database, but I'm not 100% sure.  I'll dig into that and find out.

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemSynchronizationController.java
jeffjohnson9046: The `JobExecutionListener` implementation will check to see if an instance of the job is already running, although I'm not 100% sure it goes all the way to the database vs. checking in memory on the server itself (see my earlier comment on the `JobExecutionListener`.  I sincerely hope it works that way, but hope doesn't get one very far in software development.

I definitely see some value in responding with a 4xx response to the caller if the job is already running (e.g. a 409 - Conflict, 422 - unprocessable or 423 - locked).  I think that's a good enhancement - I'll create a JIRA for it.

src/main/java/org/opentestsystem/ap/imrt/iis/client/ItemBankClient.java
jeffjohnson9046: Agreed.  While I'm at it, I think I'll rename this method to `getAllProjectIds()` since... y'know... that's what it does.

PR #39, opened 2018-04-06
Update the stimulus links when a stimulus is first created.

swilson11:


build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.28'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.10'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.11-SNAPSHOT'
```
swilson11: Will update once I have a common release build

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemRevisionWriterImpl.java
```diff
@@ -97,6 +98,22 @@ private void saveImrtItem(final ItemRevision itemRevision) {
     }
 
     private void saveStimulus(final ItemRevision itemRevision) {
-        stimulusRepository.save((Stimulus) itemRevision.getItem());
+        Stimulus stimulus = (Stimulus) itemRevision.getItem();
+        boolean isCreate = false;
```
jeffjohnson9046: Just curious (and sort of thinking aloud), what would you think about using an `Optional` here?  Maybe something like:

```java
final Optional<Integer> maybeItemKey = Optional.ofNullable(itemRevision.getItem().getKey());

final Stimulus savedStimulus = stimulusRepository.save(stimulus);
if (!maybeItemKey.isPresent()) {
    // Get the list of ImrtItems based on the associated stimulus id...
}
```
There might be a better/more concise way to leverage `Optional`'s `orElse` here, but I'd have to noodle with it a bit.

As it stands now the code is clear and I understand its intent, just thought this might be another way to express it.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemRevisionWriterImpl.java
```diff
@@ -97,6 +98,22 @@ private void saveImrtItem(final ItemRevision itemRevision) {
...
+        if (null == itemRevision.getItem().getKey()) {
+            // If the key has not been set yet, then we are creating this stim
+            // for the first time. We need to check and see if there are any items
+            // that already reference it, and create/update the links for them.
+            // We need to save the stimulus first, so a key gets assigned. Then
+            // we can use the key when creating/updating the link
+            isCreate = true;
+        }
+        final Stimulus savedStimulus = stimulusRepository.save(stimulus);
+        if (isCreate) {
```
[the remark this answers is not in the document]
swilson11: I can't move it - once we do the write to the database (the save call), the key will be filled in automatically within the stimulus object, which is kept in the hibernate session context. So I have to check if the key is null before I do the save. Unless I am mis-understanding your question?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemRevisionWriterImpl.java
swilson11: I like it, thanks.

PR #43, opened 2018-04-10
Add a script to execute the sync job in k8s to the Dockerfile

swilson11:


build.gradle
```diff
@@ -196,6 +196,9 @@ task dockerCopyFiles(type: Copy) {
     from("build/libs/${jar.archiveName}") {
     }
 
+    from('./sync-job.sh') {
+    }
+
     into 'build/docker'
 }
```
swilson11: This just copies the script into the build/docker directory so it is available when the docker image is built

sync-job.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+echo Executing Item Sync Job
+
+# Invoke the endpoint to kick off the sync job. The URL is the k8s service name for iis
+# TODO capture the response (jobId) to use for monitoring once that functionality is available
+curl -i "http://ap-imrt-iis-service/systemHook"
```
swilson11: I will change systemHook to sync when we are ready.

PR #45, opened 2018-04-10
IMRT-322: Set charset for unit test

jeffjohnson9046:
Set the charset for the call to `getBytes()`.  This resolves the `findBugs` issue, which complains if the charset isn't set.


src/test/java/org/opentestsystem/ap/imrt/iis/message/ItemUpdateMessageRecovererTest.java
```diff
@@ -28,7 +30,7 @@ public void setUp() {
 
     @Test
     public void shouldLogError() {
-        final byte[] mockPayload = "{ 'id' : 42 }".getBytes();
+        final byte[] mockPayload = "{ 'id' : 42 }".getBytes(StandardCharsets.UTF_8);
```
jeffjohnson9046: This one's on me; I thought running the test would be sufficient. Clearly, I was mistaken.

PR #47, opened 2018-04-12
Tweak the way we save standard ids.

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -160,7 +161,7 @@ private void createElaStandardId() {
                 // TODO when supported metadata.getPrimaryEmphasis
                 metadata.getPrimaryCommonCoreStandard())) {
             standardIdList.add(new StandardId.StandardIdBuilder()
-                    .withStandardIdKey(new StandardIdKey(baseItem, StandardIdKey.PRIMARY))
+                    .withStandardIdKey(new StandardIdKey(baseItem.getKey(), StandardIdKey.PRIMARY))
```
millerts: Made the change in common to use the key

build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.32'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.14'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.15-SNAPSHOT'
```
millerts: Will update after releasing common

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -62,6 +62,7 @@ private void createDomainModelObjects() {
         createItemGitInformation();
         createStandardIds();
         createItemLog();
+        baseItem.setStandardIds(standardIdList);
```
millerts: Set the standardIds on the item rather than saving them separately

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemRevisionWriterImpl.java
```diff
@@ -69,9 +69,6 @@ public void saveItemRevision(ItemRevision itemRevision) {
 
         itemGitInformationRepository.save(itemGitInformation);
 
-        // Save standardIds
-        standardIdRepository.save(itemRevision.getStandardIdList());
```
millerts: No longer needed.  Saving the item saves its standard ids

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
swilson11: Since we no longer need to get the list of standardIds outside this class (in itemRevisionWriterImpl), I'd recommend getting rid of the standardIdList field and it's getter. Then you can have createStandardIds() just return the list, and do baseItem.setStandardIds(createStandardIds).

swilson11:
There must be some tests affected by this change, right?

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
millerts: 👍

PR #49, opened 2018-04-13
Bugfix/only sync projects we want

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -101,6 +101,14 @@ public void init() {
         }
     }
 
+    void setGitLabApi(GitLabApi gitLabApi) {
```
millerts: I should have done this separately but I moved all the public/project private implemented methods together to easier read what is important in this class.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -341,7 +267,11 @@ private String findCreateBranch(int itemBankId, ItemGitInformation existingGitIn
             while (projectPager.hasNext()) {
                 final List<Project> projects = projectPager.next();
 
-                projectIds.addAll(projects.stream().map(Project::getId).collect(Collectors.toList()));
+                List<Integer> validProjectIds = projects.stream()
```
millerts: This is the big change.  The previous implementation took every project regardless if it was a project IIS cared about.  This is what was causing the duplicate items because developers' local branches have the same item id for items.  I didn't see an API that allows to only get projects by namespace so we filter here.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -365,6 +295,80 @@ public boolean isProjectMonitored(final int projectId) {
         }
     }
 
+    @Override
```
millerts: Just moved from above to below since these aren't implemented

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemBankItemRevision.java
```diff
@@ -60,4 +60,18 @@ public boolean isBeingCreated() {
     public String getIngestSource() {
         return ingestSource;
     }
+
+    @Override
```
millerts: This is used when we log errors on ingest.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -111,8 +111,14 @@ private ItemGitInformation syncItemRevision(ItemGitInformation existingGitInfo,
...
-        // Save to the DB
-        itemRevisionWriter.saveItemRevision(revision);
+        try {
+            // Save to the DB
+            itemRevisionWriter.saveItemRevision(revision);
+        } catch (Exception e) {
+            //Log item and project information.  By rethrowing the higher level error handler will log the Exception itself so we don't log it again here.
+            operationalEventService.serviceError(logger, null, "Item id {} failed to save.  Revision {}", revision.getItem().getId(), itemBankItemRevision);
+            throw e;
+        }
```
millerts: I'm not one to normally log twice, but by the type the Exception bubbles we only have the project id information.  One can find the item from that information but it requires hitting the gitlab API.  This will now log the itemBankRevision information and item and then rethrow.  It purposefully doesn't log the Exception because that is handled at a higher level.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
```diff
@@ -500,22 +500,28 @@ public void shouldUpdateItemOnPushEvent() {
     public void shouldGetAllProjectIds() throws GitLabApiException {
         final ProjectApi mockProjectApi = mock(ProjectApi.class);
         final Pager mockProjectPager = mock(Pager.class);
+        final int invalidProjectNumber = 77;
```
millerts: To test this is a little confusing but it seems like testing the internals of the gitlab API can be a tad difficult.  What this test now does is make the 77th item have an invalid project namespace for the configured client.  All other changes are formatting.

@jeffjohnson9046 - It doesn't look like your codestyle is the default IntelliJ.  Your indents are smaller than mine and Sue's

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -152,70 +160,6 @@ public void onProjectEvent(ProjectSystemHookEvent event) {
         }
     }
 
-    @Override
```
millerts: Moved below

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
swilson11: You could avoid the duplicate logging by creating a new Exception type with the information you want to have available, and then throw an instance of that.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
swilson11: I am not quite understanding your comment. From the code, it looks like you are filtering based on groupId. So for example on dev, we would only get projects with groupId iat-development, and we would ignore projects with groupId iat-uat. I agree this is needed, I just don't understand your comment about developers local branches?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
millerts: Done

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -116,8 +117,7 @@ private ItemGitInformation syncItemRevision(ItemGitInformation existingGitInfo,
             itemRevisionWriter.saveItemRevision(revision);
         } catch (Exception e) {
             //Log item and project information.  By rethrowing the higher level error handler will log the Exception itself so we don't log it again here.
-            operationalEventService.serviceError(logger, null, "Item id {} failed to save.  Revision {}", revision.getItem().getId(), itemBankItemRevision);
-            throw e;
+            throw new ItemIngestException(e, "Item id {} failed to save.  Revision {}", revision.getItem().getId(), itemBankItemRevision);
```
swilson11: Looks like you used String.format in your Exception, so you need to replace your {} with % formats

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
millerts: Prior to the change we were not filtering at all.  So we were getting any project on the dev gitlab instance.  Developers have local branches on gitlab development that they use for their testing.  For example, there were two projects on gitlab development that had an item.json with item id `17` on master.

For example:
iat-ray/17
iat-alex/17

This didn't filter those out.  The sync code takes the list from this call and throws it on the queue so the normal filtering in place for webhooks isn't leveraged.

src/test/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImplTest.java
jeffjohnson9046: Yep - that's definitely the case here.  I have updated my code style in IntelliJ to match what you and Sue have since I made this commit.

PR #51, opened 2018-04-18
Feature/standard

millerts:


build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.32'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.17'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.20-SNAPSHOT'
```
millerts: Will update once common is released

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -51,7 +46,7 @@ private void createDomainModelObjects() {
         findMetadata();
         createImrtItem();
         createItemGitInformation();
-        createStandardIds();
+        updateItemStandardIds();
```
millerts: I renamed this method as it now updates the item.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -130,111 +125,54 @@ private void createItemGitInformation() {
...
-                    .withClaim(metadata.getPrimaryClaim())
-                    .withContentDomain(metadata.getContentDomain())
-                    .withTarget(metadata.getPrimaryAssessmentTarget())
-                    .withEmphasis("")  //TODO - IAT required to populate this information
-                    .withCommonCoreStandard(metadata.getPrimaryCommonCoreStandard())
-                    .withUpdatedBy(itemBankItemRevision.getAuthor())
-                    .build());
-        }
+    private void updateEnglishItemStandardId(BaseItem item) {
+        item.setPrimaryClaim(StringUtils.defaultString(metadata.getPrimaryClaim()));
```
millerts: ELA items only have a primary claim.  I went back and forth if we should always just populate it it all now that these are stored differently.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -130,111 +125,54 @@ private void createItemGitInformation() {
...
+        item.setPrimaryTarget(StringUtils.defaultString(metadata.getPrimaryAssessmentTarget()));
+        item.setPrimaryContentDomain(StringUtils.defaultString(metadata.getContentDomain()));
```
millerts: @jeffjohnson9046 - IAT stores content domain two different ways even though they are the same thing.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
jeffjohnson9046: I kinda like the idea of always setting the primary claim, then setting the other claims for a MATH item.  Maybe something like:

```java
private void updateStandardIds() {
    if (metadata == null) {
        return;
    }

    final String contentDomain = metadata.getSubject().equalsIgnoreCase(ItemConstants.ItemSubject.SUBJECT_ELA)
         ? StringUtils.defaultString(metadata.getContentDomain())
         : StringUtils.defaultString(metadata.getPrimaryContentDomain());

    // set primary claim here regardless of subject, and maybe some sort of comment indicating that ELA only ever has a primary claim.
    item.setPrimaryClaim(StringUtils.defaultString(metadata.getPrimaryClaim()));
    item.setPrimaryTarget(StringUtils.defaultString(metadata.getPrimaryAssessmentTarget()));
    item.setPrimaryContentDomain(contentDomain);
  item.setPrimaryCommonCoreStandard(StringUtils.defaultString(metadata.getPrimaryCommonCoreStandard());

    if (ItemConstants.ItemSubject.SUBJECT_MATH.equalsIgnoreCase(metadata.getSubject())) {
        updateMathItemStandardId(baseItem); // <-- could probably rename this to "updateSubsequentStandardId"
    }
}
```

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
jeffjohnson9046: I'm not sure I understand this comment, but I'm not intimately familiar with the data yet.  I think that's something we can discuss offline.

jeffjohnson9046: ...and now I do.  Ignore my previous comment.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -130,111 +125,54 @@ private void createItemGitInformation() {
      * Creates the standardId objects if the associated data is present.
```
jeffjohnson9046: I think this should read `Updates the standardId fields....`

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
millerts: Got it.  I'll make this change.

PR #56, opened 2018-05-09
IMRT-353:  Feature/item sync logging

jeffjohnson9046:
[IMRT-353](https://jira.fairwaytech.com/browse/IMRT-353):  Additional logging to provide insight into what the item synchronization process is doing (and how long it's doing it for).  Also, the `batch_job_execution.exit_message` field is set in the database when the job's execution completes:

* When successful, the total number of item bank ids and the number of items that needed a webhook will be stored
* When failed, the exception and a stack trace will be recorded

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationJobExecutionListener.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationJobExecutionListener.java
@@ -9,6 +9,8 @@ import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.core.JobExecutionListener;
 import org.springframework.stereotype.Component;
 
+import java.util.Optional;
+
 /**
  * A {@link org.springframework.batch.core.JobExecutionListener} that prevents the item synchronization job from running
  * more than one instance at a time.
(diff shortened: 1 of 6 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemSynchronizationJobExecutionListener.java
```diff
@@ -12,25 +17,50 @@
...
             }
         }
     }
 
     @Override
     public void afterJob(final JobExecution jobExecution) {
-        // Release the lock
+        // Release the lock and log completion
         synchronized (jobExecution) {
             if (jobExecution.equals(activeJob)) {
```
jeffjohnson9046: This feels like kind of a lot to do within a `synchronized` block, but I wanted to make sure we're interacting with the active job in a thread-safe way.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -20,17 +26,28 @@ public ItemSynchronizationServiceImpl(final ItemBankClient itemBankClient,
     }
 
     @Override
-    public void synchronize() {
-        final List<Integer> projectIds = itemBankClient.getAllItemBankIds();
+    public ItemSynchronizationResponse synchronize() {
+        final Instant getProjectIdsStartTime = Instant.now();
+        logger.info("getting item bank ids from source control");
```
millerts: This is more a debug message?

PR #57, opened 2018-05-15
IMRT-366: Feature/handle delete item

jeffjohnson9046:
[IMRT-366](https://jira.fairwaytech.com/browse/IMRT-366):  Delete an item from the IMRT database when it is removed from GitLab (either via canceling item creation in IAT or removing the project from GitLab)

**NOTE:** Where possible, the code in this PR followed the existing coding conventions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.40'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.26'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.27-SNAPSHOT'
     compile 'org.springframework.boot:spring-boot-starter-web'
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.cloud:spring-cloud-starter-config'
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
@@ -65,6 +65,8 @@ public class GitlabClientImpl implements SystemHookListener, WebHookListener, It
     private static final Logger logger = LoggerFactory.getLogger(GitlabClientImpl.class);
     private static final String MASTER_BRANCH_NAME = "master";
     private static final String CREATE_BRANCH_PREFIX = "create_";
+    private static final String PROJECT_CREATE_EVENT = "CreateProjectSystemHookEvent";
+    private static final String PROJECT_DESTROY_EVENT = "DestroyProjectSystemHookEvent";
     private static final String ITEM_FILE_NAME = "item.json";
     private static final int MAX_ITEMS_PER_PAGE = 100;
 
@@ -153,10 +155,14 @@ public class GitlabClientImpl implements SystemHookListener, WebHookListener, It
     public void onProjectEvent(ProjectSystemHookEvent event) {
         logger.debug("onProjectEvent path {}", event.getPathWithNamespace());
         // Check to see if this is a project create event for our group
-        if (projectListener != null && event.getClass().getSimpleName().equals("CreateProjectSystemHookEvent") &&
+        if (projectListener != null && event.getClass().getSimpleName().equals(PROJECT_CREATE_EVENT) &&
                 event.getPathWithNamespace().startsWith(group.getPath())) {
             // If so, notify our listener
             projectListener.onCreateItem(event.getProjectId());
+        } else if (projectListener != null
+                && event.getClass().getSimpleName().equals(PROJECT_DESTROY_EVENT)
+                && event.getPathWithNamespace().startsWith(group.getPath())) {
+            projectListener.onDeleteItem(event.getProjectId());
         }
     }
 
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/config/MessageQueueConfiguration.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/config/MessageQueueConfiguration.java
@@ -21,7 +21,7 @@ import org.springframework.context.annotation.Configuration;
  */
 // TODO figure out how to configure the retry strategy when exceptions are thrown.
 @Configuration
-public class ItemUpdateWorkQueue {
+public class MessageQueueConfiguration {
     private static final String UPDATE_QUEUE_NAME = "item-update";
     private static final String DELETE_QUEUE_NAME = "item-delete";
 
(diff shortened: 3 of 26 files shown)


build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.40'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.26'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.27-SNAPSHOT'
```
jeffjohnson9046: Will update when AP_IMRT_Common is released.

src/main/java/org/opentestsystem/ap/imrt/iis/config/MessageQueueConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+import org.springframework.amqp.rabbit.listener.adapter.MessageListenerAdapter;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configuration for AMQP item update and delete work queues.
+ */
+// TODO figure out how to configure the retry strategy when exceptions are thrown.
+@Configuration
+public class MessageQueueConfiguration {
```
jeffjohnson9046: Since now this class configures a delete and update queue, I renamed it from `ItemUpdateWorkQueue` to `MessageQueueConfiguration`.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -153,10 +155,14 @@ public void monitorItem(int itemBankId) {
     public void onProjectEvent(ProjectSystemHookEvent event) {
         logger.debug("onProjectEvent path {}", event.getPathWithNamespace());
         // Check to see if this is a project create event for our group
-        if (projectListener != null && event.getClass().getSimpleName().equals("CreateProjectSystemHookEvent") &&
+        if (projectListener != null && event.getClass().getSimpleName().equals(PROJECT_CREATE_EVENT) &&
```
jeffjohnson9046: This could probably be simplified & made more readable.

millerts: I don't think it is possible based on singular dependency but it would be nice if we didn't need the null checks on project listener.  It will never be null in our system though the class itself allows it.

millerts: I think this could be simplified with early returns:

```
//If project listener is null we can't act on the event
if (projectListener == null) {
  return;
}

//If group path isn't one we're configured to handle then we ignore the event
if (event.getPathWithNamespace().startsWith(group.getPath()) {
  return;
}

switch event.getClass().getSimpleName():
// cases on the names
```

src/main/java/org/opentestsystem/ap/imrt/iis/message/ItemMessageListener.java
```diff
@@ -0,0 +1,5 @@
+package org.opentestsystem.ap.imrt.iis.message;
+
+public interface ItemMessageListener {
```
millerts: Javadocs please

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemDeleteNotificationHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Transactional
+    public void processItemMessage(final Integer itemBankId) {
+        logger.info("Received item delete message for {}", itemBankId);
+
+        // get a lock on the project - a project should not be deleted if someone is actively working on it
+        final long projectLockId = projectLockService.lockProject(itemBankId);
+        logger.info("Acquired lock for item id {}", itemBankId);
+        try {
+            final ItemGitInformation itemGitInformation = itemGitInformationRepository.findOneByProjectId(itemBankId)
```
millerts: Is this really an error?  For example, if we get a delete message for an item we never had do we care?  I don't think it'll happen often but what if we missed an item for whatever reason then they delete it before sync runs later that day.  Is that exceptional or can we just consider it deleted?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -11,43 +13,66 @@
...
                 itemEventListener.onCreateItem(projectId);
             }
 
             itemEventListener.onUpdateItem(projectId);
         }
 
-        return new ItemSynchronizationResponse(numberOfItemsWithoutoWebhook, allItemBankIds.size());
+        // Find out if we have any item bank ids in IMRT that do not exist in source control.  If we do, they need to be
+        // deleted from the IMRT database.
+        final List<ItemGitInformation> itemGitInformationList = itemGitInformationRepository.findAll();
```
millerts: This could be quite an expensive call.  For example, you could have 200k items in the system you'll have 200k objects.   We can leave this for now but we probably could enhance this in the future to use the paging that is in the system.  Not search but make the itemGitInformationRepository a pageable or slice repository.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -11,43 +13,66 @@
...
+        final List<Integer> itemBankIdsInImrt = itemGitInformationList.stream()
+                .map(ItemGitInformation::getProjectId)
+                .collect(Collectors.toList());
+
+        itemBankIdsInImrt.removeAll(allItemBankIdsInSourceControl);
+
+        // Delete all items that are in IMRT but not in source control - IMRT should never have more items than source
+        // control has
+        if (itemBankIdsInImrt.size() > 0) {
```
millerts: !imteBankIdsInImrt.isEmpty()?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
jeffjohnson9046: Yep - totally agree.  Furthermore, I only _really_ need the project ids/item bank ids from the `ItemGitInformation`; I don't need the whole thing.  I time-boxed myself on this & couldn't get the result I wanted, so I punted & got everything.  The bonus is that this process runs off-hours so performance isn't paramount.

I'll leave it as-is for now, but I agree this needs to be revisited.  I like the idea of a pageable repository.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
jeffjohnson9046: I was thinking along these same lines; I'll make this change.

PR #60, opened 2018-05-22
Clean up the logging so we don't get double logs in the ingest application

millerts:
Exceptions were being logged multiple time in the system.  Once in the classes impacted and again at the top of the stack.  Refactored to tighten exception handling and remove double logging.

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
@@ -58,7 +58,7 @@ public class ItemUpdateNotificationHandler {
      */
     public void processItemMessage(Integer itemBankId) {
         // TODO use operationalEventService
-        logger.info("Received update message for itembank id {}", itemBankId);
+        logger.debug("Received update message for itembank id {}", itemBankId);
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemUpdateNotificationHandler.java
```diff
@@ -58,17 +58,17 @@ public ItemUpdateNotificationHandler(ProjectLockService projectLockService, Item
...
         long lock = projectLockService.lockProject(itemBankId);
         // TODO use operationalEventService
-        logger.info("Acquired lock for item id {}", itemBankId);
+        logger.debug("Acquired lock for item id {}", itemBankId);
         try {
             // We got the lock, so process the update
             syncItemToItemBank(itemBankId);
             // TODO use operationalEventService
-            logger.info("Synchronization complete for item id {}", itemBankId);
+            logger.info("Synchronization complete for itembankid id {}", itemBankId);
```
millerts: I wonder if the logging here should be debug and instead be moved to the class that pulls messages off the queue to send for processing.

jeffjohnson9046: Yeah - I think your suggestion makes sense.

PR #62, opened 2018-06-01
IMRT-394:  Feature/gitlab client

jeffjohnson9046:
[IMRT-394](https://jira.fairwaytech.com/browse/IMRT-394):  Refactor the `GitLabClientImpl` to use a repository interface for fetching information from GitLab.

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
@@ -1,7 +1,6 @@
 package org.opentestsystem.ap.imrt.iis.client;
 
 import com.google.common.collect.Lists;
-import org.gitlab4j.api.GitLabApi;
 import org.gitlab4j.api.GitLabApiException;
 import org.gitlab4j.api.Pager;
 import org.gitlab4j.api.models.Branch;
@@ -37,6 +36,7 @@ import org.opentestsystem.ap.imrt.common.service.OperationalEventService;
 import org.opentestsystem.ap.imrt.iis.config.ItemBankProperties;
 import org.opentestsystem.ap.imrt.iis.exception.GitLabApiRuntimeException;
 import org.opentestsystem.ap.imrt.iis.model.ItemBankItemRevision;
+import org.opentestsystem.ap.imrt.iis.repository.GitLabRepository;
 import org.opentestsystem.ap.imrt.iis.util.ItemParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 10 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -72,35 +71,29 @@
...
-        try {
-            namespaces = gitLabApi.getNamespaceApi().findNamespaces(gitlabProperties.getGroup());
-        } catch (GitLabApiException e) {
-            // Cause spring init to fail
-            throw new GitLabApiRuntimeException(e, String.format("Issue connecting with itembank host %s", gitlabProperties.getHost()));
-        }
+        final List<Namespace> namespaces = gitLabRepository.findAllNamespaces(gitlabProperties.getGroup());
 
-        if(namespaces.isEmpty()) {
+        if (namespaces.isEmpty()) {
```
jeffjohnson9046: Minor format change - looks like `cmd + opt + l` added a space here.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -152,11 +141,10 @@ public void handleWebHook(HttpServletRequest request) {
...
         logger.debug("Adding project push hook id {} url {}", itemBankId, gitlabProperties.getWebhookUrl());
-        try {
-            gitLabApi.getProjectApi().addHook(itemBankId, gitlabProperties.getWebhookUrl(), true, false, false);
-        } catch (GitLabApiException e) {
-            // Generate an event
-            operationalEventService.serviceError(logger, e, "Unable to register webhook for projectId/itemBankId {}", itemBankId);
+
+        final Optional<ProjectHook> createdProjectHook = gitLabRepository.addProjectHook(itemBankId);
+        if (!createdProjectHook.isPresent()) {
+            throw new IllegalStateException(String.format("Could not create a project hook for itemBankId %d", itemBankId));
```
jeffjohnson9046: Not sure if this is the best exception to throw here - I could create one specifically for this scenario, maybe a `CreateProjectHookException` or something?

src/test/java/org/opentestsystem/ap/imrt/iis/TestConfig.java
```diff
@@ -20,7 +20,10 @@
     @Bean
     @Primary
     public ItemBankProperties getItemBankProperties() {
-        return mock(ItemBankProperties.class);
+        final ItemBankProperties mockItembankProperties = new ItemBankProperties();
+        mockItembankProperties.setHost("http://gitlab-example.com");
```
jeffjohnson9046: So I had to add this, otherwise the `GitLabApi` wouldn't stand up in the integration tests.

millerts: What if you remove this class and instead set up the itembank properties in the `src/test/resources/application.yml` file? I think this could then be removed.

src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
millerts: Now that I think through this we may need to catch the Exception you're throwing here in the listener/calling applications.  The reason being is that we don't want to indicate to gitlab that we failed to process the request because gitlab will resend?

millerts: Depending on how you decide to handle this throw I think you'll want to tweak the sync service ([here](https://github.com/SmarterApp/AP_IMRT_ItemIngestService/blob/develop/src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java#L68)) to catch so that the job doesn't fail when a hook can't be created.  

If you handle all the Exception handling in the listener I think the code in the sync can remain unchanged.

PR #63, opened 2018-06-04
Feature/add test administration data fields

millerts:


src/test/java/org/opentestsystem/ap/imrt/iis/service/ItemRevisionImplTest.java
```diff
@@ -261,13 +261,27 @@ public void mathAllCoreOnlyStandardId() throws IOException {
...
+
+        BaseItem item = itemRevision.getItem();
+
+        assertThat(item.getFormCount()).isEqualTo(4);
+        assertThat(item.getExposuresCount()).isEqualTo(600);
+        assertThat(item.getItemDifficultyQuintile()).isEqualTo(5);
+    }
+
     @Test
     public void shouldHaveOrganizationInfoMappedFromJson() throws IOException {
```
millerts: I rewrote this test because the previous one was testing that IAT's common code could deserialize the json.  The test now verifies the data is captured in the item itself.

PR #66, opened 2018-06-15
IMRT-414: Feature/item reindex

jeffjohnson9046:
[IMRT-414](https://jira.fairwaytech.com/browse/IMRT-414):  Create a Spring Batch job that will re-ingest the most recent commit for every item in the IMRT database.  The intent is to make sure the IMRT database is up-to-date with what's in GitLab (e.g. when new fields/tables are added to the IMRT database).  This is not a job that should be scheduled to run at regular intervals; typically it will run as part of a deployment.

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemIngestSingleJobExecutionListener.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemIngestSingleJobExecutionListener.java
@@ -1,7 +1,7 @@
 package org.opentestsystem.ap.imrt.iis.batch;
 
 import org.apache.commons.lang.exception.ExceptionUtils;
-import org.opentestsystem.ap.imrt.iis.model.ItemSynchronizationResponse;
+import org.opentestsystem.ap.imrt.iis.model.ItemIngestJobResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.ExitStatus;
@@ -12,14 +12,14 @@ import org.springframework.stereotype.Component;
 import java.util.Optional;
 
 /**
- * A {@link org.springframework.batch.core.JobExecutionListener} that prevents the item synchronization job from running
- * more than one instance at a time.
+ * A {@link org.springframework.batch.core.JobExecutionListener} that prevents an item ingest
+ * {@link org.springframework.batch.core.Job}s from running more than one instance at a time.
  */
 @Component
-public class ItemSynchronizationJobExecutionListener implements JobExecutionListener {
+public class ItemIngestSingleJobExecutionListener implements JobExecutionListener {
     // The active job execution, which will act as a lock.
     private JobExecution activeJob;
-    private static final Logger logger = LoggerFactory.getLogger(ItemSynchronizationJobExecutionListener.class);
+    private static final Logger logger = LoggerFactory.getLogger(ItemIngestSingleJobExecutionListener.class);
 
     @Override
     public void beforeJob(final JobExecution jobExecution) {
@@ -27,26 +27,25 @@ public class ItemSynchronizationJobExecutionListener implements JobExecutionList
         // the new job request.
         synchronized (jobExecution) {
             if (activeJob != null && activeJob.isRunning()) {
-                logger.warn("An Item Synchronization Process with Job execution id: {} is already running.  Wait for this job to complete before starting another one.", activeJob.getId());
+                logger.warn("An {} with Job execution id: {} is already running.  Wait for this job to complete before starting another one.", activeJob.getJobInstance().getJobName(), activeJob.getId());
                 jobExecution.stop();
             } else {
                 activeJob = jobExecution;
-                logger.info("Item Synchronization Process starting.  Job execution id: {}", activeJob.getId());
+                logger.info("{} starting.  Job execution id: {}", activeJob.getJobInstance().getJobName(), activeJob.getId());
             }
         }
     }
 
     @Override
     public void afterJob(final JobExecution jobExecution) {
-        // Log completion of the item synchronization process.  If the job ran into an exception, capture the stack
-        // trace.
+        // Log completion of the job.  If the job ran into an exception, capture the stack trace.
         String exitStatusMessage;
         if (jobExecution.getAllFailureExceptions().isEmpty()) {
-            final Optional<ItemSynchronizationResponse> response =
-                    Optional.ofNullable((ItemSynchronizationResponse) jobExecution.getExecutionContext().get(ItemSynchronizationResponse.EXECUTION_CONTEXT_DATA_KEY));
+            final Optional<ItemIngestJobResponse> response =
+                    Optional.ofNullable((ItemIngestJobResponse) jobExecution.getExecutionContext().get(ItemIngestJobResponse.EXECUTION_CONTEXT_DATA_KEY));
 
-            exitStatusMessage = response.map(ItemSynchronizationResponse::toString)
-                    .orElse(String.format("No response data from ItemSynchronizationService execution for Job execution id: %d.", jobExecution.getId()));
+            exitStatusMessage = response.map(ItemIngestJobResponse::toString)
+                    .orElse(String.format("No response data from %s execution for Job execution id: %d.", jobExecution.getJobInstance().getJobName(), jobExecution.getId()));
         } else {
             final String[] exceptionMessages = jobExecution.getAllFailureExceptions().stream()
                     .map(throwable -> String.format("%s: %s", throwable.getMessage(), ExceptionUtils.getStackTrace(throwable)))
@@ -56,7 +55,8 @@ public class ItemSynchronizationJobExecutionListener implements JobExecutionList
 
         jobExecution.setExitStatus(new ExitStatus(jobExecution.getExitStatus().getExitCode(), exitStatusMessage));
 
-        logger.info("Item Synchronization Process complete.  Job execution id: {}, exit status code: {}, exit message: {}",
+        logger.info("{} Job complete.  Job execution id: {}, exit status code: {}, exit message: {}",
+                jobExecution.getJobInstance().getJobName(),
                 jobExecution.getId(),
                 jobExecution.getExitStatus().getExitCode(),
                 exitStatusMessage);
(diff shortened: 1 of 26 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/batch/ItemIngestSingleJobExecutionListener.java
```diff
@@ -12,41 +12,40 @@
...
 /**
- * A {@link org.springframework.batch.core.JobExecutionListener} that prevents the item synchronization job from running
- * more than one instance at a time.
+ * A {@link org.springframework.batch.core.JobExecutionListener} that prevents an item ingest
+ * {@link org.springframework.batch.core.Job}s from running more than one instance at a time.
  */
 @Component
-public class ItemSynchronizationJobExecutionListener implements JobExecutionListener {
+public class ItemIngestSingleJobExecutionListener implements JobExecutionListener {
     // The active job execution, which will act as a lock.
```
jeffjohnson9046: Made the job execution listener more "generic", since both jobs want the same behavior (i.e. only one job running at a time).

src/main/java/org/opentestsystem/ap/imrt/iis/config/ItemIngestServiceProperties.java
```diff
@@ -10,6 +10,7 @@
     private double retryIntervalMultiplier = 2;
     private long retryMaxInterval = 5000;
     private int retryAmount = 3;
+    private int pageSize = 1000;
```
jeffjohnson9046: The GitLab API and ISS have a limit on page size, but I didn't see one in IIS.  So here it is.

src/main/java/org/opentestsystem/ap/imrt/iis/config/SpringBatchConfiguration.java
```diff
@@ -1,107 +1,36 @@
...
  * (e.g. a RESTful call)
- * 3. The same job should be able to run multiple times (i.e. it should be repeatable).  That is, the Item
- * Synchronization Process should be able to run once a day (or at some other regular interval)
+ * 3. The same job should be able to run multiple times (i.e. it should be repeatable).  That is, the job should be able
+ * to run once a day (or at some other regular interval).
  * </p>
  */
 @Configuration
 @EnableBatchProcessing
 public class SpringBatchConfiguration {
```
jeffjohnson9046: To keep this file from getting to unwieldy, I broke out the Spring Batch configuration:
* Each job has its own `@Configuration` class, w/the components that are specific to it
* The things that are shared between jobs (e.g. the execution listener, the async task executor) are here in the `SpringBatchConfiguration` class

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemIngestJobController.java
```diff
@@ -10,20 +10,29 @@
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
-public class ItemSynchronizationController {
+public class ItemIngestJobController {
```
jeffjohnson9046: I renamed this controller to be more "generic" and host both jobs.  My reasoning is that the `JobLauncher` is already here, so adding another endpoint was very simple.  I also think it's convenient to have both jobs in a single controller.

However: the `sync` and `migrate` jobs are not relate in any way, aside from the fact that they're `Job`s.  I could see a case for changing this back to `ItemSynchronizationController` and having a new `ItemMigrationController`.  I don't have a strong opinion one way or the other.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemIngestJobResponse.java
```diff
@@ -0,0 +1,25 @@
+package org.opentestsystem.ap.imrt.iis.model;
+
+/**
+ * A base class for transporting information about a {@link org.springframework.batch.core.Job}'s execution.
+ */
+public abstract class ItemIngestJobResponse {
```
jeffjohnson9046: This allows me to only have to worry about a single return type in the `JobExecutionListener#afterJob()` for logging the result of a job.  The `afterJob()` method merely calls `toString()` on the response, which then gets written to the database.  The sync job and migration job have different implementations of this class, so whatever's logged in the `batch_job_execution` table will be appropriate for the type of job.

src/main/java/org/opentestsystem/ap/imrt/iis/config/ItemIngestServiceProperties.java
millerts: I'm not sure I understand this comment.  The reason iss and gitlab have the limits are because those systems can only handle pages of this size.  So is this property used throughout the application any time we request a page of data from gitlab to allow us to configure the size of our gitlab requests consistantly?  Or is this just for your job?  If it is the latter I suggest renaming it to be more descriptive of what this represents.

src/main/java/org/opentestsystem/ap/imrt/iis/config/ItemMigrationBatchJobConfiguration.java
```diff
@@ -0,0 +1,91 @@
...
+@Configuration
+public class ItemMigrationBatchJobConfiguration {
+
+    /**
+     * Create a {@link org.springframework.batch.core.Job} that complies with the rules cited above.
+     *
+     * @param jobBuilderFactory    A {@link org.springframework.batch.core.configuration.annotation.JobBuilderFactory} for
+     *                             building the {@link org.springframework.batch.core.Job}
+     * @param itemMigrationStep    A {@link org.springframework.batch.core.Step} that executes the item
+     *                             synchronization process
```
millerts: migration not sync.  It looks like you copied/pasted the comments in this class. Please review to make sure sync isn't referenced.

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemIngestJobController.java
```diff
@@ -10,20 +10,29 @@
...
 
     @PostMapping("/sync")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     public void synchronize() throws Exception {
-        jobLauncher.run(job, new JobParameters());
+        jobLauncher.run(itemSynchronizationJob, new JobParameters());
+    }
+
+    @PostMapping("/migrate")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
```
millerts: I would really like if this controller would respond with an object that contains the status and job id.  I realize the status will be "started" for job at all times.  I find the no content response to not be very helpful for me when I call this directly during deployments.

src/main/java/org/opentestsystem/ap/imrt/iis/repository/GitLabRepositoryImpl.java
```diff
@@ -134,4 +134,16 @@ public GitLabRepositoryImpl(final GitLabApi gitLabApi,
             throw new GitLabApiRuntimeException(e, "Error attempting to get a page of projects");
         }
     }
+
+    @Override
+    public Optional<Commit> findCommit(final int itemBankId, final String commitHash) {
+        try {
+            return Optional.of(gitLabApi.getCommitsApi().getCommit(itemBankId, commitHash));
+        } catch (final GitLabApiException e) {
```
millerts: What happens when you request a commit that doesn't exist?  I think this will either throw here or will throw on 141 since you're not doing a fromNullable.  I think we should catch a 404 and return an Optional.empty().  I don't know the gitlab api well enough to know how that scenario isn't handled.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemMigrationServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+            final List<ItemGitInformation> itemGitInformationList =
+                    IterableUtils.toList(itemGitInformationPage.getContent());
+            for (final ItemGitInformation itemGitInformation : itemGitInformationList) {
+                try {
+                    itemUpdateNotificationHandler.migrateItem(itemGitInformation);
+                    numberOfItemBankIds++;
+                } catch (final Exception e) {
+                    // In this case, we want to continue processing items even if an exception is encountered.
+                    // Instead of throwing, the exception will be logged and the migration process will continue.
+                    operationalEventService.serviceError(logger,
```
millerts: This doesn't have to be in this PR but it would be helpful to have the itembank ids (or item ids) that didn't get migrated in a single list.  This will require us to hunt through the logs to find what didn't migrate, correct?

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemSynchronizationServiceImpl.java
```diff
@@ -56,7 +60,9 @@ public ItemSynchronizationResponse synchronize() {
 
         // Find out if we have any item bank ids in IMRT that do not exist in source control.  If we do, they need to be
         // deleted from the IMRT database.
-        final List<ItemGitInformation> itemGitInformationList = itemGitInformationRepository.findAll();
+        final List<ItemGitInformation> itemGitInformationList = new ArrayList<>();
+        itemGitInformationRepository.findAll().forEach(itemGitInformationList::add);
```
millerts: Can' you do `itemGitInfomationList.addAll(itemGitInformationRepository.findAll()`?

jeffjohnson9046: Unfortunately nope - it doesn't seem so.  `findAll()` returns an `Iterable<T>` and `addAll` expects a `Collection<T>`.  What I can do is change it to this:
```java
final List<ItemGitInformation> itemGitInformationList =
        IterableUtils.toList(itemGitInformationRepository.findAll());
```
... which I think is a bit cleaner/more readable.

src/main/java/org/opentestsystem/ap/imrt/iis/repository/GitLabRepositoryImpl.java
jeffjohnson9046: Good catch.  I tried the GitLab API and it returns a 404 if the commit can't be found.  The GitLab API we're using will throw the same exception, just like when we're looking for a file.  I'll add handling and a test around this.

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemMigrationServiceImpl.java
jeffjohnson9046: I thought this would be an easy change to make, but it's turning out to be a challenge.  I agree a list of errored item bank ids would be valuable to have, but I'll have to put it in another PR.

PR #69, opened 2018-06-19
Feature/imrt 416

jeffjohnson9046:
[IMRT-416](https://jira.fairwaytech.com/browse/IMRT-416):  Ingest TTS fields

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.52'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.32'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.33-SNAPSHOT'
     compile 'org.springframework.boot:spring-boot-starter-web'
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.cloud:spring-cloud-starter-config'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-common:0.3.52'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.32'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.33-SNAPSHOT'
```
jeffjohnson9046: Will update to correct release version of Common when Common has been released.

src/main/java/org/opentestsystem/ap/imrt/iis/model/ItemRevisionImpl.java
```diff
@@ -120,6 +120,12 @@ private void createImrtItem() {
             baseItem.setKey(existingImrtItem.getKey());
         }
 
+        if (item.getTextToSpeech() != null) {
+            baseItem.setVisualTTSRequired(item.getTextToSpeech().getVisualTTSRequired());
+            baseItem.setVisualTTSProvided(String.valueOf(item.getTextToSpeech().isVisualTTSProvided()));
```
millerts: Actually, why are you doing a valueOf here?

millerts: I imagine it doesn't matter but should this instead be https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#toString(boolean)

PR #77, opened 2018-07-03
Add a local log4j configuration

millerts:


src/main/resources/application.yml
```diff
@@ -12,10 +12,14 @@ logging:
   level:
     org.opentestsystem: DEBUG
     org.gitlab4j: INFO
+#  config: classpath:log4j2-spring-local.xml
```
millerts: One can uncomment this line out and the application will use this as the log4j configuration.  I'm going to add this property to the cloud config for our local dev.

PR #80, opened 2018-07-23
Feature/iat 2032

millerts:


build.gradle
```diff
@@ -49,7 +49,7 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.42'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.44-SNAPSHOT'
```
millerts: I'll change this once imrt-common is released

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemIngestJobController.java
```diff
@@ -42,4 +53,17 @@ public ItemIngestJobController(final JobLauncher jobLauncher,
                 jobExecution.getJobInstance().getJobName(),
                 jobExecution.getStatus().toString()));
     }
+
+    @PostMapping("/sync/{projectId}")
```
millerts: I can break this into its own controller if we want.  I wanted an easy way to sync/migrate a single item.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemValidationServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    @Transactional
+    public void validateItem(final int itemBankId) {
+        LOG.debug("Validating item " + itemBankId);
+
+        Optional<ItemGitInformation> maybeItemGitInformation = itemGitInformationRepository.findOneByProjectId(itemBankId);
+
+        if (!maybeItemGitInformation.isPresent()) {
+            operationalEventService.error(LOG, null, "Item git information could not be found for {}.  Ignoring." + itemBankId);
```
millerts: Do you think this should be in the service or should this throw and let the caller handle it?  This should never happen in the normal flow of code.  The only way I could see it happen is if we sync an item and the item is deleted while validation is running.

src/main/resources/application.yml
```diff
@@ -23,6 +23,7 @@ iis:
   retry-max-interval: 5000
   retry-amount: 3
   imrt-page-query-limit: 1000
+  itemValidationServiceUrl: "http://localhost:8087"
```
millerts: I personally prefer the camel case rather than the `-` separator.  Thoughts?

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemIngestJobController.java
jeffjohnson9046: Yeah... I think with the additional endpoints it makes sense to break sync and migrate jobs into their own controllers.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemValidationServiceImpl.java
jeffjohnson9046: I think this should throw and be the responsibility of the caller.  I agree the only way this could happen is if the item is deleted while the validation runs.  If that rare occurrence happens, then I think that'd be a truly exceptional case.  There's no reason to call `validateItem` for something we don't have a record of.

jeffjohnson9046: I think the call to `operationalEventService#error` should be `operationalEventService.error(LOG, null, "Item git information could not be found for {}.  Ignoring.", itemBankId);`  I think the `+` would result in a log entry of "Item git information could not be found for {}.  Ignoring.12345".

src/main/resources/application.yml
jeffjohnson9046: I prefer the `-` to camel case.

PR #85, opened 2018-07-26
Made the startup check in gitlab client more declaritive

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/client/GitlabClientImpl.java
```diff
@@ -70,43 +70,45 @@
...
+        this.itemBankProperties = itembankProperties;
         this.logger = logger;
         this.itemParser = itemParser;
         this.gitLabRepository = gitLabRepository;
+        this.groupPath = itembankProperties.getGroup();
     }
 
     @PostConstruct
     public void init() {
-        final List<Namespace> namespaces = gitLabRepository.findAllNamespaces(gitlabProperties.getGroup());
```
millerts: 95-102 is here to do fast fail on gitlab connections.  That wasn't really clear to me so I broke that out into a method that is named as such.

src/main/java/org/opentestsystem/ap/imrt/iis/repository/GitLabRepositoryImpl.java
```diff
@@ -56,7 +56,7 @@ public GitLabRepositoryImpl(final GitLabApi gitLabApi,
     @Override
     public List<Namespace> findAllNamespaces(final String projectGroupName) {
         try {
-            return gitLabApi.getNamespaceApi().findNamespaces(itemBankProperties.getGroup());
+            return gitLabApi.getNamespaceApi().findNamespaces(projectGroupName);
```
millerts: This is a bug as this method was ignoring the project group name passed in and instead using the itembank properties value.  So if someone tried to call this method with a different group other than the one configured in application yml it would receive the wrong results.

jeffjohnson9046: Good catch.

PR #87, opened 2018-07-31
Bugfix/handle rabbitmq issues

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/config/MessageQueueConfiguration.java
```diff
@@ -49,83 +47,76 @@ TopicExchange exchange() {
...
                                                                              final ItemMessageListener itemMessageListener,
                                                                              final ItemIngestServiceProperties properties,
-                                                                             final String queueName) {
+                                                                             final String queueName,
+                                                                             final RejectAndDontRequeueRecoverer itemMessageRecoverer) {
         final SimpleMessageListenerContainer container = new SimpleMessageListenerContainer();
         container.setConnectionFactory(connectionFactory);
         container.setQueueNames(queueName);
         container.setMessageListener(new MessageListenerAdapter(itemMessageListener, "handleMessage"));
-        container.setAdviceChain(RetryInterceptorBuilder.stateless()
```
millerts: Not sure why but this advice chain is the problem.  Removing this allows processing of messaging.  However, this works locally and was working until recently.

PR #90, opened 2018-08-03
Feature/iat 2039

jeffjohnson9046:
[IAT-2039](https://jira.fairwaytech.com/browse/IAT-2039):  Add fields for tracking if ASL, Braille, or Closed-Captioning files were uploaded before or after English content has been updated.


build.gradle
```diff
@@ -49,9 +49,9 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.51'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.52-SNAPSHOT'
```
jeffjohnson9046: Will update to released version of Common when appropriate

build.gradle
```diff
@@ -49,9 +49,9 @@ configurations {
...
     compile 'org.opentestsystem.ap:ap-common:0.4.8'
-    compile 'org.springframework.boot:spring-boot-starter-web'
+    compile 'org.springframework.boot:spring-boot-starter-web'
```
jeffjohnson9046: Formatting change?  Will remove when I update the Common version.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemConverterImpl.java
```diff
@@ -158,6 +161,13 @@ private void populateItemWithContentRequiredData(final Item item, final BaseItem
             baseItem.setAslAttachments(getContentAttachments(baseItem,
                     item.getAsl().getAttachments(),
                     AttachmentFileTypes.ASL));
+
+            final Instant maxUploadDate = baseItem.getAslAttachments().stream()
+                    .map(Attachment::getUploadedDate)
+                    .max(Instant::compareTo)
+                    .orElse(Instant.MIN);
+
```
jeffjohnson9046: I think there's probably a better/more efficient way to set the "uploaded prior to" (i.e. fewer loops over the attachments), but the data set is pretty small, so I don't think it'll be a huge performance impact.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemConverterImpl.java
```diff
@@ -158,6 +161,13 @@ private void populateItemWithContentRequiredData(final Item item, final BaseItem
             baseItem.setAslAttachments(getContentAttachments(baseItem,
                     item.getAsl().getAttachments(),
                     AttachmentFileTypes.ASL));
+
+            final Instant maxUploadDate = baseItem.getAslAttachments().stream()
+                    .map(Attachment::getUploadedDate)
+                    .max(Instant::compareTo)
+                    .orElse(Instant.MIN);
```
millerts: Can this ever not be found (Optional.empty())?  If not, suggest throw since that would be Exceptional rather than defaulting to Min date.  If empty is valid suggest using it rather than null for readability/understandability.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemConverterImpl.java
millerts: One suggestion if you don't want to iterate over the collections multiple times for this you could do something like:

```
private class Attachments {
  Set<Attachment> attachements;
  Instant lastUpdateDate;
}
```

You would then capture that data and return the new object from the `getContentAttachments` method.  That or you can use a `Pair` object since this is all internal.

PR #91, opened 2018-08-10
Bugfix/iat 2019

millerts:


build.gradle
```diff
@@ -49,7 +49,7 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.52'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.53-SNAPSHOT'
```
millerts: Will change after releasing.

src/main/java/org/opentestsystem/ap/imrt/iis/client/ItemEventListenerImpl.java
```diff
@@ -19,22 +20,32 @@
...
+
+    @PostConstruct
+    public void init() {
+        // Tell the client to notify us for project create and merge events
+        itemBankClient.setItemEventListener(this);
     }
 
     @Override
     public void onCreateItem(int itemBankId) {
+        if(itemIngestServiceProperties.isIgnoreWebhookCreation()) {
```
millerts: This is the change in this class.  I wanted a way to run these things locally and turn certain stuff off when we run.  For example, webhooks fail to create because Gitlab can't hit our system.  Rather than see a bunch of errors in the logs we disable it in `local-development` profile.  The default is to run.

src/main/java/org/opentestsystem/ap/imrt/iis/controller/ItemIngestJobController.java
```diff
@@ -66,4 +70,11 @@ public ItemIngestJobController(final JobLauncher jobLauncher,
 
         return ResponseEntity.ok("Migrated " + response.getNumberOfItembankIds() + " projects");
     }
+
+    @PutMapping("/delete/{projectId}")
```
millerts: I will remove these changes.  They were an accidental commit.  Would love to have it long term but need to add some security around these non Gitlab endpoints

src/main/java/org/opentestsystem/ap/imrt/iis/service/ItemBankMapper.java
```diff
@@ -9,13 +9,13 @@
 /**
  * Converts an IAT item to an Imrt Item
  */
-public interface ItemConverter {
+public interface ItemBankMapper {
```
millerts: Naming is hard.  After working with this stuff for a bit mapping seemed better to me.  Probably because Spring uses `RowMapper` in JDBC land.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemBankMapperImpl.java
```diff
@@ -41,11 +41,21 @@
  * model, with some associated git information, into IMRT domain model objects
  */
 @Component
-public final class ItemConverterImpl implements ItemConverter {
-    public ItemRevision convert(ItemBankItemRevision itemBankItemRevision, ItemGitInformation existingItemGitInformation) {
+public final class ItemBankMapperImpl implements ItemBankMapper {
+    public ItemRevision map(ItemBankItemRevision itemBankItemRevision, ItemGitInformation existingItemGitInformation) {
```
millerts: The `ItemGitInformation` class has a mapping to its Item.  Based on that we can get the `BaseItem` associated with the item bank revision.  If it is null this means it is a new item so we have to create a new Base Item.  This allows us to `update` existing items and `insert` new items rather than always relying on Hibernate to figure it out based on PK values.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemRevisionServiceImpl.java
```diff
@@ -83,38 +78,37 @@ private void saveImrtItem(final ItemRevision itemRevision) {
...
-        Optional<Integer> linkedStimulusId = itemRevision.getLinkedStimulusId();
-        if (linkedStimulusId.isPresent()) {
-            Stimulus stimulus = stimulusRepository.findById(linkedStimulusId.get());
-            if (null != stimulus) {
-                StimulusLink stimLink = new StimulusLink(imrtItem, stimulus, imrtItem.getUpdatedBy());
-                stimulusLinkRepository.save(stimLink);
-            } else {
-                logger.warn(LOG, null, "Unable to find linked stimulus in database for item id {} stimulus id {}",
-                        imrtItem.getId(), linkedStimulusId);
+        if(imrtItem.getStimulus() == null || !imrtItem.getStimulus().getId().equals(imrtItem.getAssociatedStimulusId())) {
```
millerts: Not sure why this is all green as the only thing that changed was the additional conditional here.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemRevisionServiceImpl.java
```diff
@@ -83,38 +78,37 @@ private void saveImrtItem(final ItemRevision itemRevision) {
...
-        // we can use the key when creating/updating the link
-        final Stimulus savedStimulus = stimulusRepository.save(stimulus);
-        if (!maybeItemKey.isPresent()) {
-            List<ImrtItem> imrtItems = imrtItemRepository.findAllByAssociatedStimulusId(stimulus.getId());
-            stimulusLinkRepository.save(
-                    imrtItems.stream()
-                            .map(imrtItem -> new StimulusLink(imrtItem, savedStimulus, imrtItem.getUpdatedBy()))
-                            .collect(Collectors.toList()));
+
+        if(stimulus.getKey() == null) {
```
millerts: I'll add the comment on 106 back here to explain what is going on here tweaking it a bit based on the new model.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemValidationServiceImpl.java
```diff
@@ -42,6 +51,13 @@ public void validateItem(final int itemBankId) {
         }
 
         final ItemGitInformation itemGitInformation = maybeItemGitInformation.get();
+
+        if(itemIngestServiceProperties.isIgnoreValidate()) {
```
millerts: I wanted a way to run these things locally and turn certain stuff off when we run.  This allows us to run ingest ignoring validation calls in case we don't care about it.  Rather than see a bunch of errors in the logs we disable it in `local-development` profile.  The default is to run.

src/main/resources/application.yml
```diff
@@ -6,7 +6,7 @@ spring:
       ddl-auto: validate
     database-platform: org.hibernate.dialect.PostgreSQLDialect
   datasource:
-    url: "jdbc:postgresql://localhost:5432/imrt"
+    url: "jdbc:postgresql://localhost:5432/imrt_temp"
```
millerts: I will revert this change.

jeffjohnson9046: Probably want to change this back to `imrt`.

PR #92, opened 2018-08-10
Bugfix/iat 2151

jeffjohnson9046:
[IAT-2151](https://jira.fairwaytech.com/browse/IAT-2151):
* Add endpoint that allows the caller to abandon a job (stop the job if it's in a running state then mark it as abandoned).
* Additional feedback in `JobExecutionResponse`

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/SpringBatchJobManagementService.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/SpringBatchJobManagementService.java
@@ -2,6 +2,18 @@ package org.opentestsystem.ap.imrt.iis.service;
 
 import org.opentestsystem.ap.imrt.iis.model.JobExecutionResponse;
 
+/**
+ * A service for interacting with Spring Batch jobs.
+ */
 public interface SpringBatchJobManagementService {
+    /**
+     * Mark a {@link org.springframework.batch.core.JobExecution} as
+     * {@link org.springframework.batch.core.BatchStatus#ABANDONED}.  If the job is in a running state, it will be
+     * stopped first.
+     *
+     * @param jobInstanceId The id of the {@link org.springframework.batch.core.JobInstance} that owns the
+     *                      {@link org.springframework.batch.core.JobExecution} to abandon.
+     * @return A {@link org.opentestsystem.ap.imrt.iis.model.JobExecutionResponse} indicating success/failure
+     */
     JobExecutionResponse abandon(final long jobInstanceId);
 }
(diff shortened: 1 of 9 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/SpringBatchJobManagementServiceImpl.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public SpringBatchJobManagementServiceImpl(final JobExplorer jobExplorer, final JobOperator jobOperator) {
+        this.jobExplorer = jobExplorer;
+        this.jobOperator = jobOperator;
+    }
+
+    @Override
+    public JobExecutionResponse abandon(final long jobInstanceId) {
+        final JobInstance jobInstance = Optional.ofNullable(jobExplorer.getJobInstance(jobInstanceId))
+                .orElseThrow(() -> new IllegalStateException(String.format("Could not find job instance for instance id %d", jobInstanceId)));
```
jeffjohnson9046: I went back & forth on this a bit; I suppose I could also throw one of the `JobAbandonmentException`s I created here.  I figured if we can't find a job for the instance id, we're in a bad way so an `IllegalStateException` is a better fit.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/SpringBatchJobManagementServiceImpl.java
```diff
@@ -0,0 +1,87 @@
...
+
+        final Set<JobExecution> runningJobExecutions = jobExplorer.findRunningJobExecutions(jobInstance.getJobName());
+
+        // If more than one running job execution is found, throw exception.  Based on how the the jobs are configured,
+        // this should never happen
+        if (runningJobExecutions.size() > 1) {
```
jeffjohnson9046: I went back & forth on this too; I don't see a way for the Spring Batch to get into this state aside from a code/programming bug.

src/main/java/org/opentestsystem/ap/imrt/iis/service/SpringBatchJobManagementService.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.ap.imrt.iis.service;
+
+import org.opentestsystem.ap.imrt.iis.model.JobExecutionResponse;
+
+/**
+ * A service for interacting with Spring Batch jobs.
+ */
+public interface SpringBatchJobManagementService {
```
jeffjohnson9046: Long term, I think I'd like the controller to call methods on this interface to start the `sync` and/or `migrate` jobs.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/SpringBatchJobManagementServiceImpl.java
millerts: Isn't this something a user passes in?  If so isn't a NotFoundException more appropriate?  Or something that would trigger a 404?

PR #94, opened 2018-08-13
Change mapping to flatten attachments

millerts:


src/main/java/org/opentestsystem/ap/imrt/iis/repository/impl/ItemFormRepositoryImpl.java
```diff
@@ -0,0 +1,28 @@
...
+        this.formRepository = formRepository;
+        this.entityManager = entityManager;
+    }
+
+    @Override
+    public void deleteAllFormsByItemId(final BaseItem item) {
+        formRepository.deleteAllByItem(item);
+
+        //This is required since there is no way to uniquely identify Forms
+        entityManager.flush();
```
millerts: I don't love this solution but this allows us to delete all the forms before reinserting them.  Since we can identify forms we have to recreate them every single time.

PR #97, opened 2018-08-16
Add validation calls passing in branch name.  Also only migrate once …

millerts:
…on sync

### DIFF
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemIngestServiceImpl.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemIngestServiceImpl.java
@@ -89,7 +89,10 @@ public class ItemIngestServiceImpl implements ItemIngestService {
             }
 
             final ItemBankItemRevision itemBankItemRevision = itemBankClient.getItemBankItemRevision(itemGitInformation);
-            return Optional.of(syncItemRevision(itemGitInformation, itemBankItemRevision));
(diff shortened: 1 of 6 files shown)


src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemIngestServiceImpl.java
```diff
@@ -126,6 +129,11 @@ public ItemIngestServiceImpl(ProjectLockService projectLockService,
             existingGitInfo = syncItemRevision(existingGitInfo, revision);
         }
 
+        if(existingGitInfo != null) {
```
millerts: This change makes it so we only validate the latest version.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemIngestServiceImpl.java
```diff
@@ -89,7 +89,10 @@ public ItemIngestServiceImpl(ProjectLockService projectLockService,
             }
 
             final ItemBankItemRevision itemBankItemRevision = itemBankClient.getItemBankItemRevision(itemGitInformation);
-            return Optional.of(syncItemRevision(itemGitInformation, itemBankItemRevision));
+            final ItemGitInformation updatedGitInformation = syncItemRevision(itemGitInformation, itemBankItemRevision);
+            itemWorkQueueService.sendItemValidationMessage(updatedGitInformation.getProjectId());
```
millerts: This change makes it so we only validate the latest version.

PR #102, opened 2018-08-29
Feature/iat 2049

jeffjohnson9046:
[IAT-2049](https://jira.fairwaytech.com/browse/IAT-2049):  Add fields for validation results count fields.

### Additional Notes
After testing a query that would get the counts from the `validation_result` table for each of the severity levels, I decided it would be better to store the counts whenever the item is submitted to the validation service.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,7 +49,7 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.60'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.62-SNAPSHOT'
     compile 'org.opentestsystem.ap:ap-common:0.4.15'
     compile 'org.springframework.boot:spring-boot-starter-web'
     compile 'org.springframework.boot:spring-boot-starter-actuator'
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.opentestsystem.ap:ap-imrt-common:0.1.62-SNAPSHOT'
-    compile 'org.opentestsystem.ap:ap-common:0.4.17'
+    compile 'org.opentestsystem.ap:ap-common:0.4.20'
     compile 'org.springframework.boot:spring-boot-starter-web'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -49,8 +49,8 @@ configurations {
 dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
-    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.61'
-    compile 'org.opentestsystem.ap:ap-common:0.4.17'
+    compile 'org.opentestsystem.ap:ap-imrt-common:0.1.62-SNAPSHOT'
```
jeffjohnson9046: Will update to released version of `ap-imrt-common` when appropriate.

src/main/java/org/opentestsystem/ap/imrt/iis/service/impl/ItemValidationServiceImpl.java
```diff
@@ -62,8 +83,30 @@ public void validateItem(final int itemBankId) {
...
+        final BaseItem baseItem = itemGitInformation.getItem();
+        final Map<String, List<ValidationResult>> validationResultsGroupedBySeverity = results.stream()
+                .collect(groupingBy(result -> result.getSeverity().toLowerCase()));
+
+        baseItem.setSevereValidationResultCount(validationResultsGroupedBySeverity.getOrDefault("severe", Collections.emptyList()).size());
+        baseItem.setDegradedValidationResultCount(validationResultsGroupedBySeverity.getOrDefault("degraded", Collections.emptyList()).size());
+        baseItem.setTolerableValidationResultCount(validationResultsGroupedBySeverity.getOrDefault("tolerable", Collections.emptyList()).size());
+        baseItem.setBenignValidationResultCount(validationResultsGroupedBySeverity.getOrDefault("benign", Collections.emptyList()).size());
+
+        final long lockId = projectLockService.lockProject(itemGitInformation.getProjectId());
```
jeffjohnson9046: I don't really like having to include the `ProjectLockService`, `ImrtItemRepository` and the `StimulusRepository` here - there's a fair amount of duplication between this and the `ItemRevisionService#save` method.  Unfortunately, to leverage that method I'd have to get an `ItemRevision`, which involves using the `GitlabClientImpl` to get all the data I need to build one.  Also, the `ItemRevisionService#save` method would also create another log entry in the `item_log` table, which we don't want.  

The _other_ thing is we aren't really dealing with a revision here; we're just updating some counts for an item, which has nothing to do with GitLab.

I suppose another alternative would be to build a `BaseItemRepository` that handles some of this, but that seemed like a bigger change.

PR #105, opened 2018-08-31
Feature/iat 2239

millerts:


build.gradle
```diff
@@ -131,6 +131,14 @@ task unitTest (type: Test) {
...
 }
 
+task bootRunLocal(type: org.springframework.boot.gradle.run.BootRunTask) {
+    doFirst() {
+        main = project.mainClassName
+        classpath = sourceSets.main.runtimeClasspath
+        systemProperty "spring.profiles.active", "local-development"
+    }
+}
+
```
millerts: The above isn't part of this ticket, but I didn't think it was necessary to create a separate PR.  Running `gw bootRun` Spring will use the default profile.  In our application that has graylog configured and after a while it'll crash because it can't find graylog running locally.  This is a convenient way to run local spring boot via Gradle/Spring's bootRun.  The command is now `./gradlew bootRunLocal`

PR #135, opened 2019-08-02
Set logging during bootstrap phase when running in local-development …

wtritch:
…mode

Right now we have slow startups when running in local-development since the default logging configuration is used during the bootstrap phase.


wtritch:
We may also want to consider pulling the greylog logging configuration out of services and into a collator that runs in kubernetes.  Ideally our services would always just log to STDOUT.

PR #138, opened 2019-10-06
Feature/iat 3308

millerts:


src/main/resources/application.sql.yml
```diff
@@ -23,7 +23,7 @@ validation-result:
   
   delete: >-
     DELETE FROM validation_result
-    WHERE key IN (:key)
+    WHERE key IN (:keys)
```
wtritch: nit: I've run into issues where this falls apart if `:keys` is empty or null.  You may want something like:
```
WHERE (COALESCE(:keys) IS NULL OR key IN (:keys))
```
Or protect it from Java with a fast-fail/fast-return if `keys` is empty

millerts: Good call.  I'll add an IllegalArgumentException throw to the repo

PR #140, opened 2019-11-20
Feature/iat 3475

bdrainer:
This was a fun one.  Ingest is the only service, so far, that needs to call other services as a system user.  That required putting an interceptor before the outbound calls to validation and IMS.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,10 +41,6 @@ targetCompatibility = 1.8
 configurations {
     // Exclude logback, we will use log4j
     compile.exclude module: 'spring-boot-starter-logging'
-    // This is pulled in from ap-common, we don't want security
-    // on our endpoints, at least for now
-    compile.exclude module: 'spring-boot-starter-security'
-
     // Avoid having to list the findbugs annotations twice
     testCompile.extendsFrom compileOnly
 }
@@ -55,6 +51,7 @@ dependencies {
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     compile 'org.hibernate:hibernate-java8:5.0.12.Final'
     compile 'org.springframework.boot:spring-boot-starter-web'
+    compile 'org.springframework.boot:spring-boot-starter-security'
     compile 'org.springframework.boot:spring-boot-starter-actuator'
     compile 'org.springframework.cloud:spring-cloud-starter-config'
     compile 'org.springframework.boot:spring-boot-starter-aop'
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceApplication.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceApplication.java
@@ -1,12 +1,9 @@
 package org.opentestsystem.ap.imrt.iis;
 
-import org.opentestsystem.rdw.common.status.StatusConfiguration;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.Import;
 
 @SpringBootApplication
-@Import(StatusConfiguration.class)
 public class ItemIngestServiceApplication {
 
     static {
--- a/src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceConfiguration.java
+++ b/src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceConfiguration.java
@@ -1,7 +1,13 @@
 package org.opentestsystem.ap.imrt.iis;
 
+import org.opentestsystem.ap.common.config.ItemBankProperties;
+import org.opentestsystem.ap.common.config.JwtSecurityConfiguration;
+import org.opentestsystem.ap.common.restclient.AppendSystemAuthenticationHeaderInterceptor;
+import org.opentestsystem.ap.common.security.jwt.JwtService;
 import org.opentestsystem.rdw.common.status.StatusConfiguration;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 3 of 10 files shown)


build.gradle
```diff
@@ -41,10 +41,6 @@ targetCompatibility = 1.8
 configurations {
     // Exclude logback, we will use log4j
     compile.exclude module: 'spring-boot-starter-logging'
-    // This is pulled in from ap-common, we don't want security
-    // on our endpoints, at least for now
-    compile.exclude module: 'spring-boot-starter-security'
-
```
bdrainer: This pulls in some needed class to the classpath.  Primarily Jackson modules used by the JwtService object mapper.

src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceApplication.java
```diff
@@ -1,12 +1,9 @@
 package org.opentestsystem.ap.imrt.iis;
 
-import org.opentestsystem.rdw.common.status.StatusConfiguration;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.Import;
 
 @SpringBootApplication
-@Import(StatusConfiguration.class)
```
bdrainer: I'm moving all configuration from the application to the primary configuration class.   You'll see this in other PRs.

src/main/java/org/opentestsystem/ap/imrt/iis/ItemIngestServiceConfiguration.java
```diff
@@ -13,14 +20,25 @@
...
+    StatusConfiguration.class
+})
+@EnableConfigurationProperties({ItemBankProperties.class})
 public class ItemIngestServiceConfiguration {
+
+    @Bean
+    public AppendSystemAuthenticationHeaderInterceptor appendSystemAuthenticationHeaderInterceptor(
+        JwtService jwtService, ItemBankProperties itemBankProperties) {
+        return new AppendSystemAuthenticationHeaderInterceptor(itemBankProperties, jwtService);
+    }
```
bdrainer: Here is the inceptor that puts the system user in the outbound request

src/main/java/org/opentestsystem/ap/imrt/iis/repository/impl/ItemTabulatorRepositoryImpl.java
```diff
@@ -27,9 +28,10 @@
 
     public ItemTabulatorRepositoryImpl(final OperationalEventService operationalEventService,
                                        final RestTemplateBuilder restTemplateBuilder,
-                                       final ItemIngestServiceProperties itemIngestServiceProperties) {
+                                       final ItemIngestServiceProperties itemIngestServiceProperties,
+                                       final AppendSystemAuthenticationHeaderInterceptor interceptor) {
         this.operationalEventService = operationalEventService;
-        this.restTemplate = restTemplateBuilder.build();
+        this.restTemplate = restTemplateBuilder.additionalInterceptors(interceptor).build();
```
bdrainer: Here is where the interceptor is set on the RestTemplate

src/main/resources/application.yml
```diff
@@ -2,8 +2,14 @@
 # when using the remote configuration service.
 management:
   security:
-    enabled: "false"
+    enabled: false
 
+security:
+  basic:
+    enabled: false
```
bdrainer: security is set to disabled for ingest, we only needed security for some classes used by the jwt service's object mapper

src/main/resources/application.yml
```diff
@@ -37,6 +43,9 @@ iis:
   item-management-service-url: "http://localhost:8081"
 
 itembank:
+  systemUser:
+    userName: "iat-imrt-ingest@smarterbalanced.org"
+    fullName: "Item Ingest"
```
bdrainer: This information is used when putting the system user in the authentication header.  Not having it is fine, but it would use the default `IAT System User`.  I'll confirm this is set in QA, UAT, Stage, and Prod

src/main/java/org/opentestsystem/ap/imrt/iis/repository/impl/ItemTabulatorRepositoryImpl.java
millerts: I feel like this should be in a Configuration class and that the restTemplate can be injected rather than the builder.

src/main/java/org/opentestsystem/ap/imrt/iis/repository/impl/TimsItemRepository.java
```diff
@@ -25,16 +25,19 @@
...
 
+    private final ItemIngestServiceProperties itemIngestServiceProperties;
 
-    @Autowired
-    public TimsItemRepository(final RestTemplateBuilder restTemplateBuilder, final ItemIngestServiceProperties itemIngestServiceProperties) {
-        this.restTemplate = restTemplateBuilder.build();
+    public TimsItemRepository(final RestTemplateBuilder restTemplateBuilder,
+                              final ItemIngestServiceProperties itemIngestServiceProperties,
+                              final AppendSystemAuthenticationHeaderInterceptor interceptor) {
+        this.restTemplate = restTemplateBuilder.additionalInterceptors(interceptor).build();
```
millerts: Same comment.  I feel like this configuration should be moved to a configuration class and all of these can take in the RestTemplate as an injected dependency.

src/main/java/org/opentestsystem/ap/imrt/iis/repository/impl/ItemTabulatorRepositoryImpl.java
bdrainer: I like it.  I'll make that change.

PR #142, opened 2019-12-06
Fix hibernate item json deserialization

wtritch:
We needed to bump the hibernate-types library version to allow us to configure the hibernate json ObjectMapper instance.


src/main/resources/hibernate-types.properties
```diff
@@ -0,0 +1 @@
+hibernate.types.jackson.object.mapper=org.opentestsystem.ap.common.model.ModelObjectMapper
```
wtritch: Note that this will perform versioning when de-serializing item.json

PR #143, opened 2019-12-17
IAT-3545 - Add new "Test Category" fields to TIMS

alex-ponce:
Added 3 new metadata fields

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ group=org.opentestsystem.ap
 
 version=0.5.49-SNAPSHOT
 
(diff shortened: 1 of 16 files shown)


gradle.properties
```diff
@@ -11,7 +11,7 @@ group=org.opentestsystem.ap
 
 version=0.5.49-SNAPSHOT
 
-commonVersion=0.5.75
+commonVersion=0.5.76-SNAPSHOT
```
alex-ponce: Will replace with common version once released

